PR #3, opened 2015-07-02
Bugfix. See DokanOperationProxy

twd2:
:)


Liryna:
Hi,

You are right we need to change the param name but your change is still misleading.
We should have the same params named as https://github.com/dokan-dev/dokany/blob/22cf77c32594ec74e12038e9bc470e57524b8e9e/dokan/volume.c#L26.

Liryna:
Thank you !

PR #6, opened 2015-08-18
Unit test project for Dokan.NET

viciousviper:
A starting point for all those who have been wondering how Dokan.NET forwards the various calls to methods declared on `System.IO.FileSystemInfo` subtypes to an implementation of the `DokanNet.IDokanOperations` interface.

Presently still a work in progress, but most simple scenarios are covered. See the test methods in `DriveInfoTest`, `DirectoryInfoTest`, and `FileInfoTest` for details.

**Remarks**
* This project requires .NET 4.6
* The test `DirectoryInfoTest.Delete_WhereRecurseIsTrueAndDirectoryIsEmpty_CallsApiCorrectly()` persistently fails. This may be an error in Dokan.NET/DokanY.


Liryna:
Thank you for your contribution ! :trophy:
This is a very good initiative !
When you say it is still in progress, do you mean that I should wait before testing ?
Anyway I have to update my environment to make the test :D

viciousviper:
You're very welcome.
When I first came across the Dokan project I felt that this was a hidden diamond - actually a bit of a rough diamond seeing how there is virtually no detailed documentation of its behavior apart from the stories and myths spread across various coding portals.
After some fiddling with the interface I got fed up with the situation and decided it was time for a more systematic approach. So there we are.
Concerning the "work in progress" aspect
* I think the basic wiring for the test suite is pretty much in place.
* What is still lacking is the number of test cases because so far I concentrated on happy day scenarios (`DokanResult.Success`) and simple path structures.
* Of the presently 41 tests there are only 8 bad day cases (`!= DokanResult.Success`).
* There are no tests yet for deep path structures, large volume transfers, or timing aspects.
* There are probably two or three places where a bit of refactoring could improve code coherence.
Apart from that - by all means do test the code. I'm looking forward to any feedback.

viciousviper:
Ok, so I'd say the unit test project has matured sufficiently to be ready for inclusion into Dokan.NET. Please give it a spin if you can spare the time.
I've implemented at least one test for each of the relevant properties/methods of the System.IO classes `FileInfo`, `DirectoryInfo`, and `DriveInfo` for a total of 61 test cases of which only one fails inexcpliably.
All in all the Dokan calls into the .NET proxy look pretty consistent. So far I've found only three likely glitches/errors in the call sequences:
* In `FileInfo.MoveTo()`, `DirectoryInfo.MoveTo()`, and `FileInfo.Replace()` the `Cleanup()`/`CloseFile()` sequence of the target files/directories is called **twice** (with `DokanFileInfo.IsDirectory` in both boolean states `true` and `false`). Apparently the driver doesn't distinguish between files and directories and calls both cleanups, just to be safe.
* Again in `FileInfo.MoveTo()`, `DirectoryInfo.MoveTo()`, and `FileInfo.Replace()` there is an irritating call to `IDokanOperations.OpenDirectory()` with a **blank** filename (`string.Empty`) and no associated cleanup calls. This is redundant and should be removed from the driver.
* In the recursive delete call `DirectoryInfo.Delete(true)` I have not found a way to handle the case of recursively deleting an empy directory. When `IDokanOperations.OpenDirectory()` returns `DokanResult.Success` on the directory marked out for deletion, the subsequent processing calls into `System.IO.Directory.DeleteHelper()` - which apparently cannot cope with a situation where the collection of child `FileSystemInfo` objects returned by `IDokanOperations.FindFiles()` is empty.

BTW, I have no idea what appveyor's problem is in compiling the project. Maybe you have access to a detailed failure report. The temporary merge conflict has been resolved with checkin 3345454.

Liryna:
Tomorrow for sure !
And I will look at the issue you have report.
Thank you again for this amazing PR :heart:

viciousviper:
One more thing. I'm afraid I have to report a regression from Dokan 0.7.3 RC4 (special build) to 0.7.4.
Two of the tests in `DriveInfoTests` - `GetDriveFormat_CallsApiCorrectly()` and `GetVolumeLabel_CallsApiCorrectly()` - used to pass and now silently fail somewhere outside of the .NET code.
I can see the `OpenDirectory()` and `GetVolumeInformation()` methods, respectively, being called. Both return `DokanResult.Success`. Execution then fails ouside the scope of the .NET debugger with the following message:
> The active Test Run was aborted because the execution process exited unexpectedly. To investigate further, enable local crash dumps either at the machine level or for process te.processhost.managed.exe. Go to more details: http://go.microsoft.com/fwlink/?linkid=232477
I'd definitely like a second opinion on this situation before I try to investigate further.

Liryna:
I am currently testing your PR. I was able to run a part of the test but since the merge to master, I always get the same error as you described in your last post.
I agree it come from dokany. I just wanted to reproduct the regression :cry:

Could you just check if everything was merged well and still work as you described from now ?

Also, why moving DokanNet from 4.0 to 4.5.2 ?

Liryna:
I have added the Unit test to Appveyor.

https://ci.appveyor.com/project/Liryna/dokan-dotnet/build/0.7.4-13/job/ok00k22dsl644okt

Liryna:
@viciousviper I face the same error `WhereRecurseIsTrueAndDirectoryIsEmpty_CallsApiCorrectly` as you.

I made a test with mirror.c: mirror.exe /d /s /r C:\Users /l m
and a C# test
```
var sut = new DirectoryInfo("M:\\Dir");
sut.Delete(true);
```
The delete work.
Here is the mirror log:
```
###Create 0138
   CreateDisposition 0x00000001
CreateFile : C:\Users\Dir
  AccountName: liryna, DomainName: DESKTOP-RBPTIQA
        OPEN_EXISTING
        ShareMode = 0x7
        FILE_SHARE_READ
        FILE_SHARE_WRITE
        FILE_SHARE_DELETE
        AccessMode = 0x80
        FILE_READ_ATTRIBUTES
        FlagsAndAttributes = 0x2000000
        FILE_FLAG_BACKUP_SEMANTICS

CreateFile status = 0
###GetFileInfo 0138
GetFileInfo : C:\Users\Dir
        GetFileInformationByHandle success, file size = 0

###Cleanup 0138
Cleanup: C:\Users\Dir

###Close 0138
Close: C:\Users\Dir

###Create 0139
OpenDirectory : C:\Users\Dir

CreateFile status = 0
###FindFiles 0139
FindFiles :C:\Users\Dir\*
        FindFiles return 2 entries in C:\Users\Dir\*

index from 0
FileMatch? : . (*,0,0)
  =>return single entry
index to 1
index from 1
FileMatch? : . (*,1,0)
FileMatch? : .. (*,1,1)
  =>return
index to 2
index from 2
FileMatch? : . (*,2,0)
FileMatch? : .. (*,2,1)
  STATUS_NO_MORE_FILES
###Cleanup 0139
Cleanup: C:\Users\Dir

###Close 0139
Close: C:\Users\Dir

###Create 0140
OpenDirectory : C:\Users\Dir

CreateFile status = 0
###GetFileInfo 0140
GetFileInfo : C:\Users\Dir
        GetFileInformationByHandle success, file size = 0

###SetFileInfo 0140  13
DeleteDirectory C:\Users\Dir
  dispositionInfo->DeleteFile = 1
###Cleanup 0140
Cleanup: C:\Users\Dir

        DeleteOnClose
  DeleteDirectory success

###Close 0140
Close: C:\Users\Dir
```

FindFiles return 2 entries in C:\Users\Dir\*

viciousviper:
@ Liryna

> Could you just check if everything was merged well and still work as you described ?

Will do that - might take me until tomorrow, though.

> Also, why moving DokanNet from 4.0 to 4.5.2 ?

Seemed like a good idea seeing how long 4.5.2 has been available. If you need to keep compatibility with older runtimes, then 4.0 is the way to go.

> Edit:
> With last DokanNet
> Dokan 0.7.3-RC4:
> 
>     4 Failed
>         Delete_WhereRecurseIsTrueAndDirectoryIsEmpty_CallsApiCorrectly
>         MoveTo_CallsApiCorrectly
>         MoveTo_CallsApiCorrectly
>         Replace_CallsApiCorrectly

The lower three likely fail due to the workaround I had to inclode for 0.7.4 (the irritating call to `OpenDirectory("")`. This call was not present in 0.7.3-RC4.

>     51 Passed
>     8 Not Run tests
>         GetDriveFormat_CallsApiCorrectly
>         GetDriveType_CallsApiCorrectly
>         GetIsReady_CallsApiCorrectly
>         GetName_CallsApiCorrectly
>         GetRootDirectory_CallsApiCorrectly
>         GetTotalFreeSpace_CallsApiCorrectly
>         GetTotalSize_CallsApiCorrectly
>         GetVolumeLabel_CallsApiCorrectly
> 
> Dokan 0.7.4:
> 
>     1 Failed
>         Delete_WhereRecurseIsTrueAndDirectoryIsEmpty_CallsApiCorrectly
>     54 Passed
>     8 Not Run tests
>         GetDriveFormat_CallsApiCorrectly
>         GetDriveType_CallsApiCorrectly
>         GetIsReady_CallsApiCorrectly
>         GetName_CallsApiCorrectly
>         GetRootDirectory_CallsApiCorrectly
>         GetTotalFreeSpace_CallsApiCorrectly
>         GetTotalSize_CallsApiCorrectly
>         GetVolumeLabel_CallsApiCorrectly
> 
> Not Run tests = The active Test Run was aborted because the execution process exited > > unexpectedly....
> 
> Does this results fit yours ?
> I also have a lots of difficult to run it in Debug mode. None of the test Run.
> And any idea why every DriveInfoTest fail ?

The success or failure of the DriveInfo tests is dependent on the order of test execution. I can run six of the tests successfully - all except `GetDriveFormat_CallsApiCorrectly()` and `GetVolumeLabel_CallsApiCorrectly()` - in a row and in combination with the DirectoryInfo and FileInfo tests. As soon as one of the two culprits is started, however, it silently fails and execution of the remaining tests is cut short.

Liryna:
In debug, I always have this throw and GetDriveFormat_CallsApiCorrectly is never called.

```
An unhandled exception of type 'System.Runtime.Serialization.SerializationException' occurred in Unknown Module.

Additional information: Unable to find assembly 'Moq, Version=4.2.1507.118, Culture=neutral, PublicKeyToken=69f491c39445e920'.
```

Liryna:
About `Delete_WhereRecurseIsTrueAndDirectoryIsEmpty_CallsApiCorrectly()`, FindFiles cannot return none. There is always the `.` & `..` folder.

viciousviper:
Regarding the Moq serialization error:
Yes, I've seen this, too. Appears to me like a side effect of an earlier failure, not the actual cause.
I got this exception all the time while figuring out the correct call sequences for the tests.

But for the actual test I can trace execution well in to the test method `GetDriveFormat_CallsApiCorrectly()` in the debugger. The failure occurs in line 67 when accessing `sut.DriveFormat`.

Regarding the missing `.` and `..` files:
Well found. Returning just these two entries will probably fix the test. `DirectoryInfo.DeleteHelper` actually checks for these two cases and discards them.

viciousviper:
I can confirm that adding `.` and `..` fixes the test `Delete_WhereRecurseIsTrueAndDirectoryIsEmpty_CallsApiCorrectly()`.

Liryna:
Can you PR ? :heart: 

About `GetDriveFormat_CallsApiCorrectly()`, I cannot breakpoint to line 67 :cry: can you see the value of sut and his members ?

viciousviper:
PR has been created.

When I break at DriveInfoTest line #67 the `DriveInfo` instance "sut" is displayed as "Z:\". However, I cannot inspect its properties - because the execution of DriveFormat and VolumeInformation fails :frowning:

Liryna:
Thank you
GetVolumeInformation is correctly called ?

viciousviper:
Yes, the Mock callback for GetVolumeInformation is called in

https://github.com/dokan-dev/dokan-dotnet/blob/194ddbb866188c6df1201c7aadb18b5c3027c315/DokanNet.Tests/DokanOperationsFixture.cs#L454

Then something bad happens to the return values on their way back through Dokan :cry:

viciousviper:
The debugger reports the following values in line #454 of `DokanOperationsFixture` during the execution of GetDriveFormat:

this ^ {DokanNet.Tests.DokanOperationsFixture} ^ DokanNet.Tests.DokanOperationsFixture
_volumeLabel	^ "Dokan Volume" ^ string
_features	^ CaseSensitiveSearch | CasePreservedNames | UnicodeOnDisk | PersistentAcls | SupportsRemoteStorage	^ DokanNet.FileSystemFeatures
_fileSystemName ^ "Dokan Test" ^ string
info {DokanNet.DokanFileInfo}	DokanNet.DokanFileInfo

Liryna:
Humm I not really know Moq but can you force the value somewhere in your callback ?
To be sure that Dokan get the values not through Moq.

I just made a test with the Mirror.cs and I had no issue :cry:

Liryna:
If you add a breakpoint in GetVolumeInformationProxy (DokanNet).

GetVolumeInformation throw

```
Message = "IDokanOperations.GetVolumeInformation(null, 614595217221352284, null, 
DokanNet.DokanFileInfo) invocation failed with mock behavior Strict.\r\nAll invocations on the mock 
must have a corresponding setup."
```

```
StackTrace = "   at Moq.ExtractProxyCall.HandleIntercept(ICallContext invocation, InterceptorContext 
ctx, CurrentInterceptContext localctx)\r\n   at Moq.Interceptor.Intercept(ICallContext invocation)\r\n   
at Moq.Proxy.CastleProxyFactory.Interceptor.Intercept(IInvocatio...
```

viciousviper:
I have found breakpoints to be a two-edged sword due to the interference by other processes - namely explorer.exe - which try to access the newly mounted Dokan drive as soon as it becomes visible.
However, the Moq Setup for GetVolumeInformation() is entirely generic, and the only way it could fail is an unexpected setup inside the DokanFileInfo parameter. I have not seen this with calls to GetVolumeInformation, only to file- or directory-related methods.

Anyway, when I set a breakpoint to https://github.com/dokan-dev/dokan-dotnet/blob/194ddbb866188c6df1201c7aadb18b5c3027c315/DokanNet/DokanOperationProxy.cs#L686 I get the following result:
dokannetdebug 1

viciousviper:
I can confirm that the error has nothing to do with Moq. If I remove the delegated call to Moq from `DokanOperationsFixture.Proxy.GetVolumeInformation(...)` the error still occurs.

It couldn't be a threading issue? I restrict the mount point to only a single thread in `Mounter.cs` while the Mirror example allows five threads ...

Liryna:
@viciousviper Oh yes it can be! Try to use the same value as mirror. One thread can be not enough with all calls made at the start.

Liryna:
@viciousviper I cannot check now but, are you settings the callback after mounting the device? If not, can you try to set your call back before calling mount?

viciousviper:
Simply increasing the thread count to 5 does not change anything. I just had a BSOD after changing the filesystem capabilities from NetworkDrive to RemovableDrive - didn't capture the memory dump, though.

Mounting the device after setting up the callbacks would require some refactoring of the test solution. Presently the device is mounted only once per test-suite execution in an Assembly-wide initialization method. I can try that - but not tonight.
I am also a bit uncertain about the value of such an experiment. Doesn't the Mirror example work by mounting the device first and potentially executing a GetVolumeInformation call later?

viciousviper:
Ok, I did a quick check by including a setup for `GetVolumeInformation()` in `DokanOperationsFixture.SetupMount()` which effectively does what you suggested - the callback is put in place before the actual device mount operation is called.

Still no luck, neither with 1 nor with 5 threads. The two tests still ominously fail without a catchable Exception.

Liryna:
I agree with you, it changed nothing :cry: 

I still try to understand what is the difference from DokanNetMirror and this test.
This fonction have nothing special, it should work.

Liryna:
Ok, to exclude the GetVolumeInformation C# implementation, we can try to disable SetupGetVolumeInformation for a test.
Dokan implement his own GetVolumeInformation in case GetVolumeInformation is not implemented by the user.
https://github.com/dokan-dev/dokany/blob/22cf77c32594ec74e12038e9bc470e57524b8e9e/dokan/volume.c#L92

So we should get "DOKAN" as VolumeName and "Dokan" as FS.
https://github.com/dokan-dev/dokany/blob/22cf77c32594ec74e12038e9bc470e57524b8e9e/dokan/volume.c#L54

If this is still not working, we have a issue somewhere else.

viciousviper:
Well, removing the very same line from `Dokan.cs` would avoid the error in DokanNet.Test as well - except the `GetVolumeLabel_CallsApiCorrectly()` test is bound to fail because it does not receive its expected volume label but 'Dokan Volume' instead.
But this would kinda defeat the purpose of the tests in the first place, wouldn't it?

Instead try this:
* Compile and run the mirror sample from within Visual Studio under a non-administrative user (I did in debug mode)
* open 'cmd.exe' under the same user and type `dir n:\` -> you'll get the directory listing of the mirrored C:\ directory
* now open 'cmd.exe' as local administrator and type `dir n:\`. At this point the running mirror process inside Visual Studio is terminated. After a couple of seconds an error is propagated back to the administrative 'cmd.exe' citing insufficient system resources. This is a very similar error to the one that is thrown for the two failing tests in DokanNet.Tests
* repeat the `dir n:\` thing a couple of times and you'll get a BSOD - at least I got one twice in a row - originating from `ntkrnlmp.exe ( nt!SeQueryInformationToken+2ac3 )`

What does this tell me about the internal stability of the Dokan driver? :angry:

viciousviper:
**AHA**
Now try this: I just found oht that I had bee running the test assembly with a 'Default Processor Architecture' of 'X86'. So I went ahead and switched it to 'X64' (Visual Studio Menu 'Test'/'Test settings'/'Default Processor Architecture' and guess what? I now get 63 passed unit tests. 

**WOOHOO !** :triumph:

It appears that the x64 Dokan driver does not go well with the X86 - if only for two of the tests. Why this should be the case escapes me totally.

Liryna:
Oh @viciousviper :heart: 
Running Visual Studio test runner in x64 gave me 63 passed unit tests me too !
Well found !
'Default Processor Architecture' seems to be a local configuration :(, do you have an idea how to force it by default ? (This would fix the appveyor)

I will try to see if there is a reason for this...but I am pretty surprise because the `dokan.dll` x64 was only add without really testing some month ago >< (we simply changed the build arch). I would understand if the x64 was unsable but the x86...

Do not be mad :smile: at the Dokan stability. The project was abandoned for more than 3 years by the only one maintainer and we forked it. There is a lots of work to do !

I made a try to reproduct the BSOD you gave but nothing happend for now. I will keep trying later.

Liryna:
A .runsettings can do the job but you cannot set it as default :(
```
<?xml version="1.0" encoding="utf-8"?>
<RunSettings>
    <!-- x86 | x64  
      - You can also change it from menu Test, Test Settings, Default Processor Architecture -->
    <TargetPlatform>x64</TargetPlatform>
</RunSettings>
```

Liryna:
Hey @viciousviper !
I had to do some changes in dokan-net for the next release of dokany and I faced the same issue as you with GetDriveFormat_CallsApiCorrectly() and GetVolumeLabel_CallsApiCorrectly() and moreover...GetVolumeInformation crash.

I found that it come from one of my commit :'( 
https://github.com/dokan-dev/dokan-dotnet/commit/7ac835d26f2af05d5d53ec73b335b5d75c0a31fb#diff-baf93385f8eb7095cee067baba4319c4L6
This change create a memory corruption in the C side....it is not a long that expected in the C side.

I made the fix in another branch for now, but I wanted to let you know and again, excuse me :smile:

Liryna:
@viciousviper I have also changed Cleanup and Close prototype, they now return void. I made the changes in Dokan.Test but I am not totally sure of the changes, can you take a quick look :D ?

https://github.com/dokan-dev/dokan-dotnet/commit/1812b02d9fd3cb703b72492b31a2ca3f89874ccf#diff-0f3f663a3cccd223cb4ef766aa591292R31

viciousviper:
Hi Liryna,
thanks for asking for my input on this matter.
I just pulled commit ccde031a into my import branch and ran into a blocker.
Apparently the changes wrt. NTStatus break 13 tests in the DokanNet.Tests project (including the test `CopyTo_WhereSourceIsLargeFile_CallsApiCorrectly` that I still haven't gotten around to fixing). Presently I can't say if the tests or the Dokan.NET proxy are to blame for the 12 new failures.
I'll be happy to look into the matter but it'll take me at least until tomorrow evening. If at all possible I'd ask you to delay a potential 0.8.0 release until we get the tests sorted out.

As to the C-side glitch behind `GetDriveFormat_CallsApiCorrectly()` and `GetVolumeLabel_CallsApiCorrectly()` - no need to apologize. I'm glad you were able to nail it down.

Liryna:
Yes I will make a pre-release for you !
In my side, I had only 4 failures.

I will take a look of `CopyTo_WhereSourceIsLargeFile_CallsApiCorrectly` when we will finish.

Liryna:
Here is it: https://github.com/dokan-dev/dokany/releases/tag/v0.8.0

viciousviper:
Hi Liryna,
I think appveyor runs tests in x86 - which is why we had problems with `GetDriveFormat_CallsApiCorrectly()` and `GetVolumeLabel_CallsApiCorrectly()` in the first place. Root cause seems to be that Visual Studio ignores x86/x64 settings for its test runner when you open a solution file. I tried to google up a solution a while back but didn't manage to find any solution.
Maybe we should include a pseudo test that reports back the presently running processor architecture.

Liryna:
Ok, well that's not really important :I was just curious.

viciousviper:
I can confirm that there are only 4 failed tests with dokany 0.8.0.

Liryna:
This failure are not linked with Cleanup/Close become void functions ?

viciousviper:
No, that's not a problem.
I fixed the tests to (almost) work in dokany 0.8.0 - including the previously failing `CopyTo_WhereSourceIsLargeFile_CallsApiCorrectly`.
The two newly failing tests seem to hint at a problem in the implementation of `EnumerateNamedStreams()` in dokany 0.8.0 - more details in the pull request.

PR #9, opened 2015-08-27
Additional tests for large file read and write operations

viciousviper:
Added two more tests for large file (> 64k) read and write
Changed mount options to include DebugMode and KeepAlive
Included .runsettings file
Minor refactorings and adapted CodeAnalysis ruleset


Liryna:
Thanks a lot @viciousviper :heart: !
Well found for appveyor !

About your mail, you are now part of development community of dokan too now. :smiley:

Liryna:
Sorry to bother you again but it seems that in Debug build Moq does not like that Proxy operation throw.
https://github.com/dokan-dev/dokan-dotnet/blob/194ddbb866188c6df1201c7aadb18b5c3027c315/DokanNet/DokanOperationProxy.cs#L165

To get all test passing in debug I have removed all #ifdef DEBUG and throw;.
Do you see another solution to protect Moq from this throw ?

viciousviper:
Please look into the Issue I just opened. Seems like there may be a problem in the nuGet distribution of Moq. I'll look into this later today.

PR #11, opened 2015-08-27
Additional tests

viciousviper:
Implemented timing tests for file read and write operations
Excluded TestCategory "Timing" from AppVeyor
Added tests for recursive directory listing and access to non-root directories

### DIFF
--- a/DokanNet.Tests/DirectoryInfoTest.cs
+++ b/DokanNet.Tests/DirectoryInfoTest.cs
@@ -24,8 +24,7 @@ namespace DokanNet.Tests
             DokanOperationsFixture.ClearInstance();
         }
 
-        [TestMethod]
-        [TestCategory(TestCategories.Success)]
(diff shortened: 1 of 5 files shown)


Liryna:
Thank you !
We begin to have a big amount of test with your contributions !

viciousviper:
You're welcome - and I'm not without self-interest in this matter. It just helps understanding the behavior of the Dokan library a little better.
After the last checkin I'm about done with the range of tests that I envisioned up front. There is one more aspect I'd like to explore, which is the behavior of the `context` property in  class `DokanFileInfo` across separate API calls.
Otherwise I'd hope that the test suite will lead to the elimination of the several anomalies it uncovered - namely the inexlicable calls to `OpenDirectory(string.Empty, ...` in some `DirectoryInfo` methods. The fixes for these will have to be made in the dokany project, however. Do you have a suggestion for the best way to track the progress of these dokany fixes?

Liryna:
#12

PR #13, opened 2015-08-28
Additional tests for non-root paths

viciousviper:


Liryna:
As I understand, `string.Empty` will have to be replace by `'\'`for the next release of dokany ?
https://github.com/dokan-dev/dokany/commit/b5e7a2c9ad624e2510dbf995c5bde0a1f6acc75e

viciousviper:
Yes, that's right.
I'll be happy to patch the tests to accept `\` instead of `string.Empty` for the next version of *dokany*.

PR #14, opened 2015-08-29
 Implement tests for the passing of context info via the DokanInfo.Context property

viciousviper:
Additional tests for FileInfo read/write operations using a context object passed via the DokanInfo.Context property
Adjusted version number of test assembly to main assembly


Liryna:
:+1:

PR #15, opened 2015-08-30
Added tests for flushing, locking, and unlocking of a FileStream

viciousviper:
during FileInfo read/write operations


Maxhy:
Merged. Thanks!

PR #18, opened 2015-09-05
Avoid throwing .NET Exceptions back to Dokan driver

viciousviper:
DokanNet
  `DokanOperationsProxy` now returns `DokanResult.ExecptionInService` except for rethrowing the .NET Exception to the Dokan driver in all build targets
  Activation of tracing switched to define TRACE instead of DEBUG
  Minor fixes to glitches in tracing and minor reformatting

Sample\\Mirror
  Minor fixes to glitches in Mirror sample
  Added tracing to Mirror sample

DokanNet.Tests:
  Fixed tests for directory and subdirectory creation
  Improved handling of Exceptions thrown in tests

### DIFF
--- a/DokanNet.Tests/DirectoryInfoTest.cs
+++ b/DokanNet.Tests/DirectoryInfoTest.cs
@@ -742,9 +742,9 @@ namespace DokanNet.Tests
             // WARNING: This is probably an error in the Dokan driver!
             fixture.SetupOpenDirectoryWithoutCleanup(string.Empty);
             fixture.SetupMoveFile(path, destinationPath, false);
-            fixture.SetupCleanupFile(destinationPath, isDirectory: true);
             // WARNING: This is probably an error in the Dokan driver!
-            fixture.SetupCleanupFile(destinationPath/* This call is probably redundant. */);
+            fixture.SetupCloseFile(destinationPath, /* This call is probably redundant. */isDirectory: true);
+            fixture.SetupCleanupFile(destinationPath);
 #endif
 
             var sut = new DirectoryInfo(DokanOperationsFixture.DirectoryName.AsDriveBasedPath());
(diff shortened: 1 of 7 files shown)


viciousviper:
I've taken the liberty to fix some errors in DokanNet and the Mirror sample and apply some reformatting.
This is still no refactoring to the Mirror sample - which it could use in one or three places. Basic functionality remains as before.

viciousviper:
@Liryna
Please check for yourself if the PR really works with C#5/VS2013.
I did my best to avoid any new C#6 constructs but did not actually cross check on an older Visual Studio / pure .NET 4.5.

Liryna:
Ah ! Thats a good PR as always :smiley: !

I cannot acces my dev env for now :cry: I will make a test monday/tuesday with VS2013.

Does `DokanResult.ExceptionInService` behave well with dokan driver ?

viciousviper:
The Dokan driver seems to have no problem with `DokanResult.ExceptionInService`. As I wrote in the other issue thread I did not suffer any BSODs after changing from rethrown Exceptions to `DokanResult`.
I look forward to an updated nuGet package - and even more to an updated Dokan driver :smile:.
You probably already saw that the nuGet package of Dokan.NET is already at 1.0.7.0 which is a mismatch to dokany 0.7.4. So will it be Dokan.NET 1.**0.8.0** to match dokany 0.8.0?

Liryna:
It will come in time :smiley:
But yes I think it is going to be 1.0.8.0.
It is more @Maxhy who manage this usually.

Liryna:
Ok for me !
Thanks as always :smiley: 

TestCategory(TestCategories.Manual)
CopyTo_WhereSourceIsLargeFile_CallsApiCorrectly
Does this test have problem to run ? Do you know the reason ?

viciousviper:
I'm still working on this test. So far I haven't been able to make it pass although I don't see anything wrong in the Mock setups. If you like you are welcome to take a look at the situation. It is one of the more advanced scenarios where I'm not sure that I'm passing all the required information to the Dokan driver. Copying files works as long as the entire file buffer is returned on the first call to `ReadFile()`. This test returns a lower `bytesRead` count than requested by the driver, however, and expects `ReadFile()` to be called several times with increased offsets.
The other test that failed in the AppVeyor build passes every time on my local machine.

PR #23, opened 2015-09-28
Update tests to match dokany 0.8.0

viciousviper:
Hi Liryna,

I've been able to get all but two of the tests in the `DokanNet.Tests` project to succeed against dokany 0.8.0 with this commit.

The two test which still fail are `FileInfoTest.Replace_WhereParentIsRoot_CallsApiCorrectly()` and `FileInfoTest.Replace_WhereParentIsDirectory_CallsApiCorrectly()` and I have reason to believe that their persistent failure is actually caused by an error in the implementation of `EnumerateNamedStreams(...)` in dokany:
  - Returning an error `NtStatus` value such as `NtStatus.Error` from `EnumerateNamedStreams(...)` to the Dokan driver results in either an `AccessViolationException` or in varying errors of either missing or excessive Mock setups during test execution.
  - Returning `NtStatus.Success` with `streamName = string.Empty` and `streamSize = 0` leads to an endless repetition of calls into `EnumerateNamedStreams(...)` by the Dokan driver although I found reference documentation here (http://www.flexhex.com/docs/articles/alternate-streams.phtml) stating that a streamSize of zero should indicate no (more) alternate streams on a file.

A similar situation occurs with the five `FileInfoTest.CopyTo*(...)` tests. The Dokan driver doesn't take kindly to receiving an error `NtStatus` from its calls to `EnumerateNamedStreams(...)`, but at least I don't get an endless recursion for `NtStatus.Success` / `streamName = string.Empty` / `streamSize = 0`.

Please look into this, otherwise the tests seem to be ok.


viciousviper:
The AppVeyour timeout is obviously caused by the same endless repetition of calls to `EnumerateAlternateStreams(...)` :frowning:

Liryna:
You need to return an error when no Streams are found with EnumerateAlternateStreams.
When you return success, it mean that there is other streams to read and he will continue to call you.

Liryna:
Yes the dokany code of EnumerateAlternateStreams is not made to return no Streams. I will fix it and update appveyor to see.

viciousviper:
There's one more thing that strikes me as strange. The file system capabilities returned by the call to `GetVolumeInformation(...)` do not include `FileSystemFeatures.NamedStreams`.
Why, then, does the Dokan driver call `EnumerateNamedStreams` on the DokanNet.Tests filesytem in the first place?

Liryna:
@Maxhy will better answer this than me

Liryna:
From What I see in this dokany commit https://github.com/dokan-dev/dokany/commit/8989997fcbef13b5aeadd38cacfb495739449fc2

If https://github.com/dokan-dev/dokan-dotnet/pull/23/files#diff-4f7a718f776f61f8a001f9cf4178ad89R1077 Success is changed to NotImplemented.
It is possible that it work.

If you can make a quick edit on github @viciousviper , could you just change it to see if appveyor can run it :D ?

viciousviper:
Seems to work better.
However, Appveyor now keeps the test jobs open until it runs into the 40 minute timeout.
Did anything change in the way unmounting works - or can you think of another reason why the test jobs suddenly don't finish any more?

Liryna:
I think EnumerateNamedStreams is still flooding but we do not see :(
I will make a test on my env when I can.

viciousviper:
Release build just passed the tests - but it still doesn't terminate when it should.

Liryna:
On my test env, vstest.discoveryengine.exe continue to run and keep using it.
I seen a difference by removing the NetworkDrive flag...but this have to be check.

Liryna:
Ok after review, DokanMain does not return so the test does not stop.
This is a dokany issue.

I merge this and will try to correct dokany.

PR #32, opened 2015-10-09
Fixed EnumerateNamedStreams to give a fake result.

shlatchz:


Liryna:
I was more expecting a real implementation :cry:

shlatchz:
Oh.. so I didn't understand you right :\

Liryna:
Here is some documentation about Alternate Data Streams in NTFS
http://blogs.technet.com/b/askcore/archive/2013/03/24/alternate-data-streams-in-ntfs.aspx

PR #33, opened 2015-10-10
Add support to EnumerateNamedStreams for Mirror Example

shlatchz:
For some reason, rawStreamNameLength is 1/2 its real size.
Because of that, I had to put the stream name twice.


Liryna:
Looks already better ! but try to use nuget packages.

shlatchz:
I have used a nuget package..
What should I include in my pull exactly? It's my first time and all.. I wanna do it right :\

Liryna:
@shlatchz No problem, the best would be to have only one commit for each pull request. Just commit the c# files and package.config file and do not include binaries.

PR #34, opened 2015-10-10
Add support for EnumerateNamedStreams at DokanNetMirror

shlatchz:


shlatchz:
I noticed that you can only enumerate existing streams. You cannot add streams or read their content.
Will it be possible in the future?

Liryna:
@shlatchz This seems to be better :+1:
I will look into it tomorrow. 
About your question, createfile is going to be call with the name of the file (as usual) but with the stream name also. So read and write operation  will be performed like any other stream.

shlatchz:
Cool! :+1:

Liryna:
@shlatchz But I don't think the C# wrapper handle it :p

shlatchz:
Why not?

Liryna:
For exemple, FileStream does not support it. https://github.com/dokan-dev/dokan-dotnet/blob/master/sample/DokanNetMirror/Mirror.cs#L134
and natively C# does not too.
http://stackoverflow.com/questions/604960/ntfs-alternate-data-streams-net
It mean that even if we enumerate them, for now we will not be able to answer correctly to the create/read/writer operations

Also, I am not a big fan of using a commercial library in the mirror. I know that there is not a lot of solutions but better to not add dependancy that we cannot manage in the futur.
And why did you double the length of the name in the proxy O_o ?
https://github.com/dokan-dev/dokan-dotnet/pull/34/files#diff-0f3f663a3cccd223cb4ef766aa591292L968

shlatchz:
I doubled it because when I tested it, I got only half of the stream's name...
For some reason rawStreamNameLength didn't hold the full length of the stream's name.
That's why I doubled the value.

Liryna:
Yes correct, this is related to a dokany issue.
I will change the implementation to not have to give the stream size.

Liryna:
The length issue has been fixed with https://github.com/dokan-dev/dokany/commit/7799f883acd08d75dbcf185d6dd53581f845a3bb
Otherwise, I cannot accept this PR in that state since it will create more instability to have a half implementation of the alternate streams.

PR #35, opened 2015-10-11
Expose mount option for drive-specific timeout

viciousviper:
Expose the mount config option for a drive-specific timeout value (see https://github.com/dokan-dev/dokany/commit/3a7968b55be3323b3bb89eba8ac62a7345f6ac66) in the DokanNet wrapper.


viciousviper:
AppVeyor is throwing a fit again - I have no idea why the entire test run fails.
All tests succeed on my local build. Go figure :frowning:
Sorry for the largish diff - it appears that line endings were inconsistent in the preexisting `Dokan.cs`.

Liryna:
Hi, @viciousviper ,
Thanks for the PR, it is right that this option was missing :smile: 

Could you just rebase your commit to have only one ? 4 commit for one change is a lot :(

viciousviper:
Canceled due to commit ordering issues

PR #36, opened 2015-10-15
Expose mount option for drive-specific timeout

viciousviper:
Expose the mount config option for a drive-specific timeout value (see https://github.com/dokan-dev/dokany/commit/3a7968b55be3323b3bb89eba8ac62a7345f6ac66) in the DokanNet wrapper.

### DIFF
--- a/DokanNet/Dokan.cs
+++ b/DokanNet/Dokan.cs
@@ -1,3 +1,4 @@
+using System;
 using DokanNet.Native;
 
 namespace DokanNet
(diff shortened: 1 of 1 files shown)


Liryna:
Hey @viciousviper !

I can accept this PR if you want :smiley: !

viciousviper:
Hi again,
sorry for the prolonged silence - I took a couple of days off to catch the last sunny days for this fall. :sunglasses:

I'm having some mixed feelings about the present stability of dokan-dotnet/dokany as a whole. The latest few test runs on AppVeyor look rather catastrophic and I experienced a regression on my local build machine as well where the FileInfo.Replace* tests suddenly broke without any apparent reason in the .NET code. This is actually why I pulled the trigger on this PR - couldn't make sense of the erratic errors popping up seemingly from within dokany.
Would it possibly make sense to wrap the progress made at the dokany end of things into another preview build? As it stands now I'd rather temporarily demote the FileInfo.Replace* tests to manual than do the trial and error dance required on dokany 0.8.0RC2 to get them working again.

Liryna:
Hi @viciousviper ,

No problem, It is important to relax :tropical_drink: when you can.

Do you have an easy case where FileInfo.Replace fail ? that I can reproduce and analyse ?
AppVeyor is currently failing because the dokan-net is only compatible with dokany & not with last RC.
EnumerateNamedStreams has been changed to FindStreams.

PR #43, opened 2015-11-20
Fix unit tests for dokany 0.8.0-RC3

viciousviper:
- rewrite former calls to former CreateDirectory/OpenDirectory methods
- fix FileAccess, FileShare, and FileOptions parameters in multiple tests


viciousviper:
Unit tests run perfectly on my local environment (83 tests **passed**) for checkin https://github.com/viciousviper/dokan-dotnet/commit/7c50dd02f98159a35f0f0dc9c6dc291d031d5691 on dokany 0.8.0-RC3. In Appveyor only 72 tests pass while 11 tests fail.
The problems in Appveyor seem to arise from the `CloseFile()` method being called with a value of both of `true` and `false` in its `IsDirectory` parameter.

@Liryna
Could you perhaps verify the state of affairs on a local environment and give me feedback before I start taking things apart to get Appveyor back on track?

Liryna:
All 85 tests passed successfully on my side ! Good job !
It is strange that Appveyor has another behaviour :(

Liryna:
@viciousviper Have you been able to look for fixing appveyor ?

viciousviper:
I'm sorry to say that I'm experiencing even more severe problems than that.
For the last three days I have not been able to call the .NET based mount operation via `NativeMethods.DokanMain(...)`. Every time I try I get a 'System.ExecutionEngineException'  with additional information

    FatalExecutionEngineError was detected
    Message: The runtime has encountered a fatal error. The address of the error was
    at 0xee52657f, on thread 0x2b4c. The error code is 0xc0000005. This error may be
    a bug in the CLR or in the unsafe or non-verifiable portions of user code. Common
    sources of this bug include user marshaling errors for COM-interop or PInvoke, 
    which may corrupt the stack.

This happens for both the Dokan.NET unit tests and the .NET based Mirror sample. The native code Mirror sample works as before, however.
This is on Windows 8.1 x64 with the x64 dokan-0.8.0-RC3 driver. I already reinstalled the dokan driver a couple of times but that didn't change anything. Eventlog dutifully reports the .NET corruption but does not shed any more light on the reasons behind the failure.

I'm a bit at a loss here.

Liryna:
I made some heavy test on my windows and impossible to get an error :(
I will try to test it on Windows 8.1 x64 as you in the next days.
Are you sure there is no dokan.dll remaining somewhere that could be used instead of the release ? or anything that could make your environment unstable ?

I am pretty surprised you have an issue with DokanMain. It have not changed at all since last release. Does the C mirror is running correctly ?

viciousviper:
Hi Liryna,
thanks for double-checking.
It is entirely possible that my problems have nothing to do with dokany as such. I keep my machine updated for all the software packages that I use on a regular basis, so there have been a number of changes between the last time I was able to run Dokan.NET - 11 days ago - and today. Most notably I deployed the GhostDoc extension in my VS2015 and pulled in the usual MS updates as well as updating NVidia drivers to 359.00.
However, I am not aware of anything that could likely cause this kind of persistent error. None of my other .NET solutions seems to be affected so there definitely is a connection to dokany. It couldn't be some kind of x64 alignment mismatch again?

There are precisely two instances of dokan.dll on my system drive, one in `C:\Windows\system32`, the other in `C:\Windows\SysWOW64`. Cryptographic hashes (CRC32 + SHA1) match the contents of the package https://github.com/dokan-dev/dokany/releases/download/v0.8.0-RC3/DokanInstall_0.8.0-RC3.exe.

To answer your bottommost question: The **native** Mirror sample works as always, whereas the **.NET** Mirror sample crashes pretty much in the same way as the Dokan.NET unit tests on my machine.

I'll try running this setup on a different machine over the next couple of days and report back.

viciousviper:
@Liryna: It is just as you expected. Everything allright on a different machine - this one running Windows 8 x64. Need to get my main rig back in order. I'm very curious to find out what the reason is behind all this hassle.

Liryna:
:+1: Glad to hear it !

viciousviper:
Test results of the latest checkin are slightly better that before - only 5 and 6 failing tests on Release and Debug builds, respectively. My local tests are all green as before.
Looks to me as though there are race conditions in the Appveyor test execution, especially regarding calls to `CloseFile(...)`. Some of the failures could be explained by `CloseFile(...)` invocations bleeding over from a previous test, and in one instance an expected invocation of `CloseFile(...)` did not occur.
@Liryna: Do you know by chance if Appveyor runs MSTest tests on multiple parallel threads or tries to be clever with async test execution?

P.S.: I got my rig up and running again. Maybe it was a lucky coincidence that VS2015Upd1 / .NET 4.6.1 came along yesterday - or it could have been a lingering instance of `Dokan.dll` in the old Regist**o**ryFS directory #-)

viciousviper:
There.
I'll go about replacing the timeouts with a more robust solution later.

Liryna:
Ahh that's perfect :smiley: ! Yes the sleep are not the best solution...but well, we are back on green !

Thank you again @viciousviper for your time !

PR #45, opened 2015-12-07
Improve single core thread handling

viciousviper:
- Replace Thread.Wait(...) with Thread.Yield(...) to allow more efficient synchronizationof alternate threads in single core environments.
- Detect pending files count underflow (< 0)


Liryna:
Thank you @viciousviper as alway ! :heart:

viciousviper:
You're very welcome. Happy to see that a new major version is drawing close.

PR #48, opened 2015-12-25
Tests for Overlapped file read/write operations

viciousviper:


Liryna:
@viciousviper should I start review your PR ?

viciousviper:
@Liryna: Yes please.
I'm a bit irritated about the single failing (old) test in AppVeyor, but otherwise the new test class for Overlapped file access via `ReadFileEx()`/`WriteFileEx()` is operational.
It proves, too, that multithreaded access is handled correctly by Dokany/Dokan.Net in spite of the errors produced by Rambalac's other test suite. I'm still having to track down the reasons for those errors.

I'd be especially interested to know your opinion on the diverging `dwNumberOfBytesTransferred` values provided in the calls to the `ReadFileEx()`/`WriteFileEx()` callbacks. Uncommenting the `Assert` statements against `outputs[i].BytesTransferred` makes all four tests in **OverlappedTest.cs** fail consistently because the values of `BytesTransferred` for all chunks are erroneously reported to equal the size of the last (smaller) chunk.
My suspicion is that I'm hitting a limitation of **Moq** here but I haven't been quite able to pinpoint the root cause yet.

Liryna:
Hi @viciousviper,

All test are working good !!!

Do you have a idea for fixing appveyor ?

About dwNumberOfBytesTransferred, I have the same behaviour as you reported. It seems that this parameter is a source of issue...http://stackoverflow.com/questions/13518045/writefileex-completion-routine-succeeds-but-bytes-transferred-is-incorrect
Does the content of read buffer is correct when dwNumberOfBytesTransferred is wrong ?

Thank again for all your work :smiley: !

viciousviper:
Hi Liryna,

unfortunately I don't have any idea why `CopyTo_WhereSourceIsEmpty_CallsApiCorrectly()` fails in AppVeyor at this time. It is unchanged from the time of Dokan.NET 1.0.8.0 when all tests passed. Looks like on of the calls bleeds over from one of the other tests

I've checked the stackoverflow reference you suggested - alas the solution presented there (including `FILE_FLAG_NO_BUFFERING` in the options parameter of `CreateFile()`) does not change anything wrt. `dwNumberOfBytesTransferred`. I suggest we put this detail issue to rest for the time being and I'll keep looking for a solution.
The actual copying of the `byte[]` buffer is not effected by this apparent kernel32 API bug.

Liryna:
Ok so if buffer content is ok we will do this :)

Thank you for your clear answer.

PR #49, opened 2015-12-28
Please ignore

viciousviper:
Debugging code for unmatched Mock call to CreateFile(...) in the test **FileInfoTests.CopyTo_WhereSourceIsEmpty_CallsApiCorrectly**


viciousviper:
@Liryna: There we have it. :cold_sweat:

Please take a look at the diff for **FileInfoTests.cs** between commits 0909ece9ef6fbb0865d3902c6e59fc8b92ef0493 and 81f9a0be7cf38eaf6f31c1db7a061f6ce2d37fee for details.

For some reason only known to itself, Appveyor makes an additional call to
`CreateFile[2] ("\DestinationFile.txe", [ReadData, WriteData, AppendData, ReadExtendedAttributes, WriteExtendedAttributes, ReadAttributes, WriteAttributes, Delete, ReadPermissions, ChangePermissions, Synchronize], [None], OpenOrCreate, [SequentialScan], [Normal], DokanFileInfo {<null>, False, False, False, False, ..., False, False})
`
after the call to
`CreateFile[2] ("\DestinationFile.txe", [ReadData, WriteData, AppendData, ReadExtendedAttributes, WriteExtendedAttributes, ReadAttributes, WriteAttributes, Delete, ReadPermissions, ChangePermissions, Synchronize], [None], CreateNew, [SequentialScan], [Normal], DokanFileInfo {<null>, False, False, False, False, ..., False, False})`
that is not present in local MSTest execution. :see_no_evil:
The only obvious difference is `FileOpen.CreateNew` instead of `FileOpen.OpenOrCreate`.

I am totally at a loss here. :sob: Does Appveyor run on a grossly different NTFS or OS version (I'm still on Win8.1) or what else could be wrong here?

Liryna:
@viciousviper Oh :( as I know they are using azure virtual machine with Windows server 2012 r2
https://www.appveyor.com/docs/installed-software

It is possible that it is Windows server that react differently.

viciousviper:
Results from the experiments in this PR have been transferred to PR #51
--> closed this PR

PR #50, opened 2016-01-09
Master

DaveWeath:
I did not implement FindStreams, I set Files to null and returned NotImplemented. The assert then caught me even though I did not return Success.
DaveWeath


Maxhy:
This doesn't looks like a pull request. I think you made a mistake here.

Liryna:
@DaveWeath , I see your issue. I will move the assert after the check.

Liryna:
@DaveWeath , Fixed with https://github.com/dokan-dev/dokan-dotnet/commit/50b8b12f8cbd64b0eaa3d8a75728dd3e3ab0c6ad
Thank you for reporting the issue !

PR #51, opened 2016-01-10
 	Fix for failing test 'CopyTo_WhereSourceIsEmpty_CallsApiCorrectly' in AppVeyor CI build

viciousviper:
**NOTE:** This fix causes the above mentioned test to pass when called in sequence with other tests. It will, however, FAIL when executed individually.
Further analysis is required.


Liryna:
appveyor dont really like it :cry:

viciousviper:
Hi Liryna,
good to see that you're online.
I'm definitely running into a situation here that I can't solve on my own.

- The test `CopyTo_WhereSourceIsEmpty_CallsApiCorrect` that I was trying to fix in the first place still passes on my local environment both when executed individually and when executen as part of the entire suite. It fails on AppVeyour as you just saw with a redundant call raised by the Dokan driver.
- Among the set of four tests dealing with *Overlapped* reads/writes, the two *Read...* tests have lately turned unstable. They produce errors in the transferred byte[] buffers at varying positions. Apparently my earlier report of Overlapped I/O working was premature.
- Adapting the *MoveTo.../Replace...*-tests to the additional debug assertion introduced in your commit 50b8b12 now leads to a new instability in the *Replace* tests. When I execute them individually in their present form (commit 194171a) they produce an error of type "Wait cycles exceeded", meaning that some configured `Close(...)` \ `Cleanup(...)` calls are not raised, or an entirely senseless call to an unsatisfied Mock execution - see example below. When I execute them together - or with any number of additional tests from the suite - I get a variation of errors oscillating between the `AccessViolationException` also reported by AppVeyor, an `IOException` indicating an illegal file name, or again calls to unsatisfied Mock setups for entirely strange configurations.

		Testname:	Replace_WhereParentIsRoot_CallsApiCorrectly
		Testresult:	Failure
		...
		Found Mock invocations without corresponding setups
		...
		FindStreams[2]	("\DestinationDir\DestinationFile.txe", out [0], DokanFileInfo {<null>, False, False, False, False, 7020, False, False})
		GetFileInformation[2]	("\Dir\File.ext", out [FileInformation {\Dir\File.ext, [Normal], 10.01.2016 16:36:35, 10.01.2016 16:36:35, 10.01.2016 16:36:35, 0}], DokanFileInfo {<null>, False, False, False, False, 7020, False, False})
		FindStreams[2]	("\Dir\File.ext", out [0], DokanFileInfo {<null>, False, False, False, False, 7020, False, False})
		CreateFile ("\DestinationDir\DestinationFile.txe\", DokanFileInfo {<null>, False, False, False, False, 7020, False, False}) -> **MockException**: IDokanOperations.CreateFile("\DestinationDir\DestinationFile.txe\", ReadData, ReadExtendedAttributes, ReadAttributes, ReadPermissions, Synchronize, ReadWrite, Delete, Open, SequentialScan, 0, DokanNet.DokanFileInfo) invocation failed with mock behavior Strict.
		All invocations on the mock must have a corresponding setup.
		Cleanup[2]	("\DestinationDir\DestinationFile.txe", DokanFileInfo {<null>, False, False, False, False, 7020, False, False})
		CloseFile[1]	("\DestinationDir\DestinationFile.txe", DokanFileInfo {<null>, False, False, False, False, 7020, False, False})
		Cleanup[1]	("\Dir\File.ext", DokanFileInfo {<null>, False, False, False, False, 7020, False, False})
		CloseFile[0]	("\Dir\File.ext", DokanFileInfo {<null>, False, False, False, False, 7020, False, False})

Note the strange trailing `SOH` character behind "\DestinationDir\DestinationFile.txe\" - looks like a memory corruption to me.

Liryna:
Hi @viciousviper ,

My fault about the Assert, I made it too quick. Fixed with https://github.com/dokan-dev/dokan-dotnet/commit/15afd0b0ba2cb0367eda86b02ae11290626ab918

When I run the test on my Win10, I have some that fail:
OpenRead_WithVariableSizes_Overlapped_CallsApiCorrectly -> Unexpected data in output...
CopyTo_WhereSourceIsEmpty_CallsApiCorrectly ->  hasUnmatchedInvocations = false :cry: 
SetAccessControl_CallsApiCorrectly -> FindFiles Assert because files = null & Cleanup 

But if I run them again one by one, all work ! Even if Visual studio crash just after....

So I don't know if it is a timing issue, or you should probably give more time between the test. Or why not using a list of driver letter to give the OS the time to remove the others ?

I go no memory corruption so I cannot say :'(

I am testing with 64bit CPU

Liryna:
Hi @viciousviper,

I have looked at the unit test like you requested (Took me a while :disappointed: )

Today, I am no longer able to run test :cry: (like appveyor). It seems that none of the test start.
The console is flood of:
```
Exception thrown: 'Moq.MockException' in Moq.dll
Exception thrown: 'Moq.MockException' in Moq.dll
Exception thrown: 'Moq.MockException' in Moq.dll
Exception thrown: 'Moq.MockException' in Moq.dll
```

I began to review the code and there is something the seems to be strange for me.
If I understand well, at the begining dokan is mount and we wait that the device is ready.
https://github.com/dokan-dev/dokan-dotnet/blob/8c1ffd21dd68efe26b6e1383a4176a89e8b4c22c/DokanNet.Tests/Mounter.cs#L16

And when it is mount, we init the dokan operation:
https://github.com/dokan-dev/dokan-dotnet/blob/47e723c4ca45e4bb837d64b7ae7d64b1b6bb3ef0/DokanNet.Tests/FileInfoTests.cs#L222

If I understood well, we kind of have a problem here. Windows 'need' to acces it to be able to say that it is 'ready' but here we wait that it is ready before we 'accept' to answer him.

Probably we should implement a minimum of calls at start for Windows ?

viciousviper:
Well, setting up the initial Mock calls after mounting the Dokan volume - all of which target the root directory - shouldn't be a problem because of https://github.com/dokan-dev/dokan-dotnet/blob/8c1ffd21dd68efe26b6e1383a4176a89e8b4c22c/DokanNet.Tests/DokanOperationsFixture.cs#L437 where `SetupMount()` is called to initialize Moq as required. At least this has always worked for me so far.

I've taken some time off Dokan.Net to pursue my other two pet projects - as you might have guessed. I've just gotten around to *almost* stabilizing the data driven ReadFileEx/WriteFileEx Overlapped tests through the Dokan driver there and I think I'll be able to get the respective Dokan.Net tests back on track by applying the insights I gained into 

If you can wait another one or two days I'll look into the current state of affairs and report back - or just fix the tests at the current stage.

viciousviper:
Closed as obsolete

PR #61, opened 2016-02-03
Add support for custom loggers

marinkobabic:
This way we can very easy support all loggers used by developers. 

Here an example how to support log4net or nLog:

- Use the implemented TraceLogger and add Log4Net/nLog listeners in the app.config file
- Create a new instance of new Logger and pass the actions for logging
- Inherit from ILogger and pass the instance of the new created class


Liryna:
Thank you @marinkobabic for the contribution !
It work well and it is clean !

Liryna:
I have reuse your ConsoleLogger for the mirror.
https://github.com/dokan-dev/dokan-dotnet/commit/1e91e154d9afb27e7468ee4ac3f4a971572d4bd0

Now it make a beautiful output for the next reports ! Thank you @marinkobabic !

FYI, Console does not need mutex. It is already thread safe.

PR #63, opened 2016-02-05
Fixed unit tests for local execution and AppVeyor

viciousviper:
Hi Liryna,

Ok, so I got the unit test suite working again - at least for local execution and for the non-Overlapped tests. It was missing a Mock setup for the new `DokanOperations.Mounted(...)` method - very likely the reason for the total failure on your local environment - and some very curious parameter combinations apparently related to ReparsePoints appear more normal now. There is also a new requirement on the `FindFiles (...)` method to return a non-null `IList<FileInformation>` that was not there before.

All this said, I have just now been able to run 96 of the 100 separate tests successfully against dokany 1.0.0-beta on my local machine. However, AppVeyor is throwing a fit again and enters an endless loop on the first test of the `DirectoryInfoTests` class. I'll look into this tomorrow.


viciousviper:
Hi Liryna,
there you go. There are still some slightly irritating anomalies - i.e. differences i behaviour between AppVeyor and local test execution - for the test `CopyTo_WhereSourceIsEmpty_CallsApiCorrectly()`, and I still haven't gotten the `ReadFileEx` / `Overlapped` tests to work reliably, but otherwise I think we're back in business with the unit test suite.
AppVeyor ran against dokany 0.8.0 as usual, my own local tests on dokany 1.0.0-beta were likewise successful.

Liryna:
Hi @viciousviper 

YEAHH :sunglasses: !
That's seems to work very well for Appveyor ! Nice work ! 

Could you rebase all your commit into one ?

viciousviper:
Replaced by PR #64

PR #66, opened 2016-02-18
Add support for custom library path

Noklafag:
`Dokan.Initialize(path)` can be used for loading Dokan DLL from custom path
By default, path is `"dokan1.dll"`


Noklafag:
GetVolumeLabel test fucks up and I don't know why

PR #73, opened 2016-03-17
Fixed unit tests for dokany/dokan-dotnet 1.0.0-RC2

viciousviper:


viciousviper:
Replaced by a squashed version in #74

PR #85, opened 2016-06-24
Taking advantage of new c#6 features

TrabacchinLuigi:
c#6 allow String interpolation wich is way more readable and nameof() so why not using it ? 
 for example mount and unmount were a little wrong ...


Liryna:
Hi @TrabacchinLuigi ,

Thank you for your contribution !

There is not reason to not use it :smiley:

TrabacchinLuigi:
this very rapid merge incourages me to do more ... :+1:

PR #86, opened 2016-06-27
Using a format provider

TrabacchinLuigi:
Created format provider, to avoid re-writing the same code over and over to format strings and having one single reliable place to control this aspects.
P.s. this commit add a reference to third party nuget package, that bridge some string interpolation features to .net4 that are available only in 4.6


Liryna:
This clean well the trace code :+1: 

Thank you @TrabacchinLuigi for your contribution again :smiley:  !

PR #91, opened 2016-07-08
Handling of DateTime.MinValue

### LINKED ISSUE #90: Handle unknown date in FileInformation
When I don't know the creation or modification date of the files, what should I set the `LastWriteTime` and `CreationTime` in `FileInformation` to?

According to [the documentation for FILE_BASIC_INFORMATION](https://msdn.microsoft.com/en-us/library/windows/hardware/ff545762(v=vs.85).aspx) I understand that I should set it to `-1`, but `DateTime` don't have any `-1`. The most logical I can think of is `DateTime.MinValue` or `default(DateTime)`, but that generate an `System.ArgumentOutOfRangeException` when `DokanOperationProxy.AddTo` use `DateTime.ToFileTime()`. 

I suggest that `DokanOperationProxy.AddTo` check for `DateTime.MinValue` and then set it to `-1`.

magol:
Fixes #90

### DIFF
--- a/DokanNet.Tests/DirectoryInfoTest.cs
+++ b/DokanNet.Tests/DirectoryInfoTest.cs
@@ -689,6 +689,73 @@ namespace DokanNet.Tests
 #endif
         }
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming",
+            "CA1702:CompoundWordsShouldBeCasedCorrectly", MessageId = "SubDirectory")]
+        [TestMethod, TestCategory(TestCategories.Success)]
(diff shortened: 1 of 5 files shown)


Liryna:
Hi @magol ,

Thank you for your contribution and specially for adding the tests ! :heart: 
The code looks good to me !

There is just appveyor that fail on the added test:
https://ci.appveyor.com/project/Liryna/dokan-dotnet/build/1.0.0-190/job/5f8apcr0lq5c6pyp#L247
https://ci.appveyor.com/project/Liryna/dokan-dotnet/build/1.0.0-190/job/5f8apcr0lq5c6pyp#L285

Liryna:
Thank you @magol  for the quick fix ! :+1:

PR #93, opened 2016-07-11
Don't format string already formatted.

magol:
If args is null or empty, the message should not be formatted.


Liryna:
Hi @magol,

Thank you for the contribution, very appreciative!

PR #98, opened 2016-08-13
Update comments, fix Doxygen problems and sync with comments in C++

magol:


Liryna:
Thank you @magol !
Thats a good and clean PR ! :smiley:

Liryna:
What did you wanted to say by:
```
Todo:
Update the second paragraph in this documentation. 
```

Otherwise, I converted doxygen most of the commands (like \see) to the xml version to try to keep it clean.
https://github.com/dokan-dev/dokan-dotnet/commit/8bf441387096764f3fdeaa4db567019c287ca153
I know that some commands does not exist in xml but at least we do our best :smile: 

I also found that link to NativeMethods does not work because it is an internal class and doxygen do not want to document them :smile: 

Keep up the good work !

magol:
Ups, it was supposed to fix the \todo before PR.
I did write the \todo because that I don't understand the second paragraph in the comment, I do not really understand the meaning of:
```C#
/// In case <see cref="FileMode.OpenOrCreate"/> and
/// <see cref="FileMode.Create"/> are opening successfully a already
/// existing file, you have to <c>SetLastError(ERROR_ALREADY_EXISTS)</c>.
```

Can you help me with that?
* What parameter is it that can have the FileModes
* Instead of `SetLastError(ERROR_ALREADY_EXISTS)`, do I return that flag?

magol:
For references to private structs, maybe we can use `\if` in some way, so that it don't be in the documentation, but still be in the code for reference.  And if we for some reason want to generate a more detailed documentation with the privates, it is possible.

Liryna:
This paragraph should be edite for ZwCreateFileProxy since there is no FileMode but `rawDesiredAccess` (that has C flags).
For IDokanOperations.CreateFile the paragraph is correct but should probably say it is the `mode`param.
`SetLastError(ERROR_ALREADY_EXISTS)` it is just before returning STATUS_SUCCESS. (the last error is retrieved later by dokan library).

I never used `\if` have you already used it on doxygen ?

Just to know, do you use doxywizard for testing the documentation ?

magol:
If I understand you right:
If parameter `mode==FileMode.OpenOrCreate | FileMode.Create` and the file specified in `fileName `already exist, you have to use the code below and then `return NtStatus.Success`
```C#
public const int ERROR_ALREADY_EXISTS = 183;
[DllImport("kernel32.dll", SetLastError = true)]
static extern void SetLastError(uint dwErrorCode);
...
SetLastError(ERROR_ALREADY_EXISTS);
```

I hope I have misunderstood it. In C #, it feels like a strange solution.

Liryna:
It is exactly :+1:

The goal is to notify that the file has not be created by the open but already was there.
Do you have another idea how this could be handle ?

magol:
Maybe we can add a new parameter to `CreateFile`, and then let `ZwCreateFile `use `SetLastError`:
```C#
NtStatus CreateFile(string fileName, FileAccess access, FileShare share, FileMode mode,
            FileOptions options, FileAttributes attributes, out bool alreadyExists, DokanFileInfo info);
```

Liryna:
This should probably be done yes.
At first I tought creating a variable just for this would be too much but compared to make the user dllimport it is much better :smile:

magol:
Great 😊

Then I think all this out parameters is bad practice in c#. 
Beter to return a structures with the returned data.   It becomes some new structures,  but I think that c# programmers feels more comfortable.
But it is a major change.

Liryna:
which other  out parameters are you pointing ?

magol:
It is a lot of function in this interface that have out parameters. But I don't know if it is a good idea to return structs instead of NtStatus.  ;)

Liryna:
Better to keep the same logic on each callback rather than have sometimes a struct because there is a out parameter.

magol:
You are right. There are more disadvantages than advantages with my proposal.

PR #100, opened 2016-08-16
Improved performance for ZwCreateFileProxy

magol:
I have test the changed implementation by run just the changed lines in a separat program.
I iterate from `0` to `0xffff ffff` to test if it was any improvement and to verify that I didn't change functionality.

With the old implementation, the loop took two hour just to go through the first `0x0521 F660` iterations.

For the new implementation, it took **7 s** to test every values from `0` to `0xffffffff`.
That's a pretty big performance improvement.

### DIFF
--- a/DokanNet/DokanOperationProxy.cs
+++ b/DokanNet/DokanOperationProxy.cs
@@ -200,24 +200,13 @@ namespace DokanNet
         {
             try
             {
-                FileOptions fileOptions = 0;
-                FileAttributes fileAttributes = 0;
                 var fileAttributesAndFlags = 0;
                 var creationDisposition = 0;
                 NativeMethods.DokanMapKernelToUserCreateFileFlags(rawFileAttributes, rawCreateOptions,
                     rawCreateDisposition, ref fileAttributesAndFlags, ref creationDisposition);
 
-                foreach (FileOptions fileOption in Enum.GetValues(typeof(FileOptions)))
-                {
-                    if (((FileOptions)(fileAttributesAndFlags & 0xffffc000) & fileOption) == fileOption)
-                        fileOptions |= fileOption;
-                }
-
-                foreach (FileAttributes fileAttribute in Enum.GetValues(typeof(FileAttributes)))
-                {
-                    if (((FileAttributes)(fileAttributesAndFlags & 0x3fff) & fileAttribute) == fileAttribute)
-                        fileAttributes |= fileAttribute;
-                }
+                var fileAttributes = (FileAttributes)(fileAttributesAndFlags & 0x00003ff7);
+                var fileOptions = (FileOptions)(fileAttributesAndFlags & 0xdc004000);
 
                 logger.Debug("CreateFileProxy : {0}", rawFileName);
                 logger.Debug("\tCreationDisposition\t{0}", (FileMode)creationDisposition);


Liryna:
The code that you propose was there some month ago but has been replaced by the code that your remove :smiley: 

The problem we faced is that the C# does not have all flag that C CreateFile has. Therefor fileAttributes & fileOptions was holding a value that does not match FileOptions & FileAttributes enums (unknow value that does not exist in the enum).

I agree that it slow the CreateFile but we should be able to clean the unknown flags.

magol:
Oh, I see
But my suggestion is not exactly the same as it was before 3242e95.

The old solution is using `0xffff c000` and `0x0000 3fff` as masks. 
In my solution, I use `0xdc00 4000 `and `0x0000 3ff7`. I have calculated them from following:

```C#
FileOptions fileOptionsMask      = FileOptions.Asynchronous  | FileOptions.DeleteOnClose   | 
                                   FileOptions.Encrypted     | FileOptions.None            | 
                                   FileOptions.RandomAccess  | FileOptions.SequentialScan  | 
                                   FileOptions.WriteThrough;

            
FileAttributes fileAttributeMask = FileAttributes.ReadOnly   | FileAttributes.Hidden       | 
                                   FileAttributes.System     | FileAttributes.Directory    | 
                                   FileAttributes.Archive    | FileAttributes.Device       | 
                                   FileAttributes.Normal     | FileAttributes.Temporary    | 
                                   FileAttributes.SparseFile | FileAttributes.ReparsePoint | 
                                   FileAttributes.Compressed | FileAttributes.Offline      | 
                                   FileAttributes.NotContentIndexed;
                               //| FileAttributes.Encrypted | IntegrityStream | NoScrubData;

```
Since both `FileOptions` and `FileAttributes` have `[FlagsAttribute]` do this ensures that only flags that are implemented in .NET remains. Flags that are not existing in .NET are masked out.

Liryna:
Ok I see ! :+1: 

Did you tested it well ?

Why did you comment `FileAttributes.Encrypted | IntegrityStream | NoScrubData` ?

magol:
As I did say in the PR, I have try to test i for all possible values on `fileAttributesAndFlags`, but it took for long time to iterate all values with the old implementation. For my implementation, it took 7 s. 
I have run all unit tests, but I haven't test it with a real driver. 

But as long as I have not thought wrong, this is mathematically exactly the same.

The reason that I have comment out `FileAttributes.Encrypted | IntegrityStream | NoScrubData` is because you in the old implementation mask this flags out with `0x3fff`.

Liryna:
:+1: Ok perfect @magol thank you for making it clear and for the contribution as always ! very appricate :)!

magol:
:-)

You asked why I commented out `FileAttributes.Encrypted | IntegrityStream | NoScrubData`
Let me ask another question; why do you mask out these flags? Is there flags not supported, or are they irrelevant in user mode?

Liryna:
To tell the truth, I have no idea :cry: 
It was there before I took the project

Do you have an idea in mind about it ?

magol:
`Encrypted` do exist in both `FileAttributes` and `FileOptionsare` with value `16384`, so the implementer may have thought it unnecessary to have the flag in two places. If you ask me, it is a wrong decision. If I as implementer want to know if this flag is set and look at `FileAttributes.Encrypted`, I get the wrong impression that it is not set. But perhaps a little late to change now.

The other flags are implemented like this:
```C#
    [ComVisible(false)] IntegrityStream = 32768,
    [ComVisible(false)] NoScrubData = 131072,
```
They are also not included in the documentation on [MSDN](https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(System.IO.FileAttributes);k(TargetFrameworkMoniker-.NETFramework,Version%3Dv4.0);k(DevLang-csharp)&rd=true), so that probably explains it.

Liryna:
Ok so only FileAttributes.Encrypted shoud be reenabled.
Can you make a PR for it ?

magol:
Ok, I can do a PR for that

And I must correct myself. I looked wrong, `IntegrityStream `and `NoScrubData `is indeed documented on [MSDN](https://msdn.microsoft.com/en-us/library/9f93fa06(v=vs.110).aspx).

**IntegrityStream**
The file or directory includes data integrity support. When this value is applied to a file, all data streams in the file have integrity support. When this value is applied to a directory, all new files and subdirectories within that directory, by default, include integrity support.

**NoScrubData**
The file or directory is excluded from the data integrity scan. When this value is applied to a directory, by default, all new files and subdirectories within that directory are excluded from data integrity.

I do not know if that is something that should be masked away or not?

Liryna:
Lets enable it if it exist :+1:

magol:
Ok :-)

magol:
Ahh, now I know why` FileAttributes.IntegrityStream` and `FileAttributes.NoScrubData` are not included.
They were introduced in the .NET Framework 4.5, and Dokan-dotnet is compiled for .NET Framework 4

.NET Framework 4 is a pretty old now, and according to [Microsoft](https://blogs.msdn.microsoft.com/dotnet/2015/12/09/support-ending-for-the-net-framework-4-4-5-and-4-5-1/), they have stopped supporting it. It is perhaps timely to release Dokan-dotnet for a newer .NET Framework version?

If you do not think we should change version at the moment, perhaps it is enough to add a comment there to remind you that the `FileAttributes ` values should be added in the future when we choose to release for a newer version. Or perhaps even add them with a #if directive so that they will automatically added when the version is updated.

Liryna:
:smile: The reason seems to be much more logic :D 
Even if .NET Framework 4 is pretty old, it is still used everywhere. I think we should keep it on this version for keeping compatibility with software not moving to .NET 4 >.

A if would be great ! This would avoid to forget it when we will move :smiley: 

Well found @magol !

magol:
Rational thought. One should not be too hasty to update if it does not lead to a radical improvement.

One alternative option may be to have both a Dokan version for both .NET 4.0 and .NET 4.5 in the NuGet package. see [here](http://brian-federici.com/blog/2013/4/17/automatically-target-multiple-frameworks-in-your-nuget-package)

> A if would be great

What would be great? That I add comment about it, or if I add a #if directive? :-)

Liryna:
Ahhh !
The best would be to have both build .NET 4.0 and .NET 4.5 in a nuget package as you proposed !
and adding #if directive for the enums :+1:

magol:
If we add .NET 4.5, how do you think we should handle the two different versions?

I have found this suggestions:
- Have two different configurations for release ("release 4.0" and "release 4.5")? Maybe allso for debug ("debug 4.0" and "debug 4.5")?
- Let the build script set .NET version during build
- Use following in the project file to build both versions when compiling in Visual Studio:
```XML
  <Target Name="AfterBuild">
    <!--Compile for .NET Framework 4.5 to
    See http://stackoverflow.com/questions/2923210/conditional-compilation-and-framework-targets -->
    <MSBuild Condition=" '$(TargetFrameworkVersion)' != 'v4.5'" Projects="$(MSBuildProjectFile)" Properties="TargetFrameworkVersion=v4.5" RunEachTargetSeparately="true" />
  </Target>
```

Liryna:
First solution would be the cleanest I think. I just hope that it would not change too much the project.
Having DokanNet build for 4.5 does not add much so if it becomes to complicate, better not to do it.

magol:
No, it does not provide any significant benefit to the code. Perhaps the improved GC in 4.5 affect performance positively. But the question is whether there is sufficient advantage to obstruct the environment.
I can do a test, and see if it is to complicated.

PR #102, opened 2016-08-18
Changelog

magol:
Please check if I've missed something.
Version 1.0.6 do I think look little too short, are there more changes before the first commit in GitHub?


Liryna:
Wow @magol :+1: 
Wasn't expecting about it ! I very appreciate the interest you are porting to dokan !!!

Thanks again for your contribution !

I think we will not be able to do more for 1.0.6...we do not have history about it :cry: (it is at this time we took the control of the project).

PR #103, opened 2016-08-18
Fix Markedown issues in CHANGELOG.md

magol:
Write a comment in CHANGELOG.ms about v1.0.6
Add some missing files to the solution.


Liryna:
Thank you @magol :+1:

PR #104, opened 2016-08-21
Error messages localized to Swedish

magol:


Liryna:
Hi @magol ,

Thank you for the translate ! :heart: 
Do you think you are able to remake the commit without the sln and csproj ?

magol:
Sorry I accidentally commit changes to the project and solution files.
I have tried to correct it now, but I'm not sure if I did the right thing with the GIT.

Liryna:
You found how to make it right !
It now looks good :+1: 

Can you also add the translate in the nuget package ?
https://github.com/dokan-dev/dokan-dotnet/blob/master/DokanNet/DokanNet.nuspec#L21

magol:
I'm getting better and better at GIT. Soon, maybe even I know what I am doing :-)

Liryna:
Perfect !

Thank you @magol for the quick fix and the contribution :)!

PR #106, opened 2016-08-31
Multi-target to .NET Framework 4.0 and 4.6

magol:
This is a experiment to add support for .NET Framework 4.6 in parallel with 4.0.
To do that, I now have two projects with almost identical content; one for 4.0 and one for 4.6.

As `StringInterpolationBridgeStrong` do exist in 4.6, but not in 4.0, I have used separate NuGet files for the projects; a feature added to [NuGet 3.2](https://docs.nuget.org/release-notes/nuget-3.2).

Pros
 - You can use Dokan in a program using .NET Framework 4.6 or newer (no conflict with `StringInterpolationBridgeStrong`)
 - You can use the new flags in `FileAttributes` introduced in .NET Framework 4.5

Cons
 - Increased complexity with two project files and multiple binaries. 

Please review and see if you think that this is a good solution.


Liryna:
Hi @magol ,

Thank you for your contribution :+1: 

It seems like appveyor is not able to build the project :'(

https://ci.appveyor.com/project/Liryna/dokan-dotnet/build/1.0.0-270/job/8yi24v9evbodt0cg

magol:
I saw it, but I do not really understand why.
Could it be that AppVeyor do not understand:
```XML
<OutputPath>bin\Release\NET$(TargetFrameworkVersion)\</OutputPath. 
```

Maybe do I have to give the path like this?
```XML
<OutputPath>bin\Release\NETv4.0\</OutputPath>
```

magol:
@Liryna I did try to change `<OutputPath>` in the projects, but it did not fix the problems. 
Do you have any idea how to fix this?

Liryna:
No idea :cry: 
I test to reproduce appveyor build that should do something like `msbuild DokanNet.sln /p:Configuration=Release /p:Platform="Any CPU"`
But wasn't able to get the same error.

It is probably not the outpath the issue but somewhere in the configuration where Release & "Any CPU" is not write correctly.

magol:
I did try to remove the platform attribute in Appleyard.yml suggested at http://help.appveyor.com/discussions/problems/4330-build-failing-with-different-project-using-different-configurations

The two dokan projects do now build.  But it is stil some problems with the sample projects.

Liryna:
well found :+1: 
Probably we can run ourself the msbuild as we do with dokany https://github.com/dokan-dev/dokany/blob/master/appveyor.yml#L189

magol:
I have now fix the NuGet problems for DokanNetMirror, and the Release configuration do now build.
But for the Debug configuration, I get a NullReferenceException for a test.

Liryna:
That looks good for me @magol ! 

Can you rebase all your commit into a single one ? (I can create a patch otherwise)

magol:
I hope that I have done the rebase correctly now (after MANY failed attempts)

Liryna:
Perfect @magol :tada: !!!
Well done for the release :+1: sorry it was not to bother you, it is just for having a clean history in the commit.

Thank again for your contribution ! very appreciate and this will be well used at the next release :heart:

magol:
No worry.
I agree that history needed to be cleaned up, and it was fun to learn :-)

PR #107, opened 2016-09-07
Clean out remnants from .NET 3.5 in the project files

magol:
In the project files it is stuf from .NET 3.5 and OneClick.

I've cleaned up everything I judge to be unnecessary or irrelevant.


Liryna:
:+1: Thank you @magol for the cleanup !

PR #108, opened 2016-09-07
Fix warnings from Code Analysis

magol:


Liryna:
Now the build is clean :smile: ! Thank you @magol !

PR #121, opened 2016-09-21
Add links in CHANGELOG.md and add missing info.

magol:


Liryna:
Thank you @magol this is way better changelog :smile:

PR #128, opened 2016-11-14
Updated mirror.cs

TrabacchinLuigi:
found a bug with drive letters case


Liryna:
Hi @TrabacchinLuigi 
What was the bug with drive letter case that you fix here ?

TrabacchinLuigi:
it was just that sometimes the case was different and so we switched to a more permissive comparison, that bit of code was not working if i was comparing "X:\\" with "x:\\" for instance

Liryna:
Ok ! Thank you @TrabacchinLuigi for the explanation and the contribution ! ❤️ !

Issue #133, opened 2017-02-03
Strange search pattern "<


TrabacchinLuigi:
So... I was experimentig with some softwares and mirror fs.
One of those got stuck in an infinite loop of trying to save a file.
It turns out that the software tries to backup the old file using this pattern: `filename.[\d{4,4}].ext`, before doing so, it asks for all the files matching that pattern, max the matches and move the old file.
it was failing because the search pattern it was using was `filename."<.ext`.
Now... that software works fine outside dokan mirror fs, substituting the `"<` with an `*` also makes it work again.
My doubt is that somewere in dokan there might be a double encoding, or an encoding error in general, any idea how to find it out ?

Rondom:
Before I try several different possibilities.... Can give a more concrete example on how to reproduce this?

TrabacchinLuigi:
Never tried c-mirror.
The software that does that is not open source, long to download and to install and very costly ... so i won't suggest trying with that, but i could try it for you, also i could try to build a demo software that behaves the same...

TrabacchinLuigi:
I've sent you an email with some details.

TrabacchinLuigi:
So i figured out how to test with mirror.c and it works, but the strange search pattern is still there

`###FindFiles 0384
FindFiles :C:\test\source\
	FindFiles return 7 entries in C:\test\source\*

index from 0
FileMatch? : New Text Document.txt (Project1"<.rvt,0,0)
FileMatch? : Project1.0001.rvt (Project1"<.rvt,0,0)
  =>return single entry
index to 1
index from 1
FileMatch? : New Text Document.txt (Project1"<.rvt,1,0)
FileMatch? : Project1.0001.rvt (Project1"<.rvt,1,0)
FileMatch? : Project1.0002.rvt (Project1"<.rvt,1,1)
  =>return
FileMatch? : Project1.0003.rvt (Project1"<.rvt,1,2)
  =>return
FileMatch? : Project1.rvt (Project1"<.rvt,1,3)
index to 3
index from 3
FileMatch? : New Text Document.txt (Project1"<.rvt,3,0)
FileMatch? : Project1.0001.rvt (Project1"<.rvt,3,0)
FileMatch? : Project1.0002.rvt (Project1"<.rvt,3,1)
FileMatch? : Project1.0003.rvt (Project1"<.rvt,3,2)
FileMatch? : Project1.rvt (Project1"<.rvt,3,3)
  STATUS_NO_MORE_FILES`

[...]

PR #134, opened 2017-02-07
SearchFilesWithPattern takes care of special chars 

TrabacchinLuigi:
Fix [issue N° 133](https://github.com/dokan-dev/dokan-dotnet/issues/133)


TrabacchinLuigi:
The fact that the second commit makes this continous integration pass makes me worry about the test machine having Z: stuck for some reason ...

Rondom:
Not sure, what is the right way to do it...

Your translation of the C-method can, no doubt, be useful to serve as a pattern for people who might want to implement their own optimized FindFilesWithPattern. For "simple" use, I think it is enough to only implement `FindFiles` and leave FindFilesWithPattern unimplemented (in that case the C-version in the C-library that you translated from, will take care of it). [`DokanIsNameInExpression` is in the C-API](https://dokan-dev.github.io/dokany-doc/html/group___dokan.html#ga2e8585b8730ee74c0cd0cadf3e8cb01a), but is not exposed in the C#-API. Instead of calling the C-version from C# and induce overhead for converting types, it might be faster to use the C#-version, but it might also be the other way round. No matter what (C# or C), it would be useful to have that method in the dokan-dotnet as well.

[See:](https://github.com/dokan-dev/dokan-dotnet/blob/80278ee221144df8ba2b0f8be6807bf3d170981c/DokanNet/IDokanOperations.cs#L144-L145)
>  If [FindFilesWithPattern] is not implemented or returns NtStatus.NotImplemented, then FindFiles is called.

But I am no expert. So let's see what @Liryna and @magol say.

TrabacchinLuigi:
i'm not even sure i've rewritten it correctly, so ... yes let's see
Now i'm more worried about having to rebase 😆

TrabacchinLuigi:
for even a simpler use we could just replace 
`" with .`
`< with *`
`> with ?`
and use the .net implementation as it was before ...

Rondom:
See the [DirectoryInfo.GetFiles Method](https://msdn.microsoft.com/en-us/library/8he88b63(v=vs.110).aspx). The semantics are different.
In `DokanIsNameInExpression` `?` and `*` have their own meaning with less legacy MS-DOS cruft attached. I have not read the .NET-method's docs in detail, but I am sure that you could let it handle some of the cases, but not all of them. I guess you can verify it by reading the docs over and over again, by looking in the .NET source or by testing manually with Process Monitor.

    * (asterisk) Matches zero or more characters.
    ? (question mark) Matches a single character.
    " DOS_DOT Matches either a period or zero characters beyond the name string.
    > DOS_QM Matches any single character or, upon encountering a period or end
        of name string, advances the expression to the end of the set of
        contiguous DOS_QMs.
    < DOS_STAR Matches zero or more characters until encountering and matching 
       the final . in the name.

TrabacchinLuigi:
I saw that, but they look similar... 
And what about the fact that the search query was actually file.*.ext and not file"<.ext ?

Rondom:
Well, similar is not the same! From the definition above, I am sure you can come up with quite some filenames where the the behaviour is *not* the same.

I do not know more details and I cannot help you further. So you are on your own. If you want to dig deeper, I think Raymond Chen might have an entry in his blog Old New Thing (really a recommended read if you are interested in the efforts MS does to preserve backwards-compatibility). One of the difficulties is that it is sometimes not entirely clear how the mapping from functions in the C-library to Kernel-syscalls is on Windows (and what kind of processing is done in between). I think you cannot use `<>"` from user space. I guess it will depend on which user space API you use, whether a question mark arrives as `?` or `>` in the kernel.  I am not an expert, but I have sometimes found myself resorting to look at the ReactOS-Source in lieu of unsuccessfully digging for documentation.

Rondom:
I got more [information about FsRtlIsNameInExpression in the ReactOS-wiki](https://www.reactos.org/wiki/FsRtlIsNameInExpression)
* [a description from the CIFS-spec about the DOS-style matching](http://ubiqx.org/cifs/rfc-draft/draft-leach-cifs-v1-spec-02.html#s3.3)
* The [Samba-project has also some testing program](https://github.com/samba-team/samba/blob/ce9b72c17abb156de8185b100f27d1ddd3c89b15/source3/torture/masktest.c) that can give some inspiration (pay attention that `*?.` can have the DOS-meaning in there.
* [ReactOS-implementation of FsRtlIsNameInExpression with some (useful!) tests](https://github.com/reactos/reactos/blob/master/rostests/kmtests/ntos_fsrtl/FsRtlExpression.c)
* The [MS-documentation](https://msdn.microsoft.com/en-us/library/ff546850(v=VS.85).aspx)

TrabacchinLuigi:
I've just copied reactOS tests, that are way more useful than the couple i wrote, and we don't pass 😁

Liryna:
Hi @TrabacchinLuigi ,

Thanks a lot for the work you are doing 👍 ! I have started to review the PR.
Does DokanIsNameInExpression_should_match_as_expected cannot be in the dokan test unit that already exist ?

TrabacchinLuigi:
Actually yes... But I tend to keep things divided by topic, and I feel this new function belong to the mirror example implementation so maybe having some specialized tests could be ok, also I think a lot of people tend to use the mirror as a base for their own projects so giving a good base to start with, seems like a good idea. Let me know.

Rondom:
I think DokanIsNameInExpression does not belong into the Mirror. It should rather be a library-function of Dokan-dotnet. As I wrote previously:
> [`DokanIsNameInExpression` is in the C-API](https://dokan-dev.github.io/dokany-doc/html/group___dokan.html#ga2e8585b8730ee74c0cd0cadf3e8cb01a), but is not exposed in the C#-API. Instead of calling the C-version from C# and induce overhead for converting types, it might be faster to use the C#-version, but it might also be the other way round. No matter what (C# or C), it would be useful to have that method in the dokan-dotnet as well.

Licensing: The ReactOS-tests are LGPLv2+. Dokan-dotnet is MIT. From my understanding we cannot simply copy-paste the code, but I have not double-checked this and I am tired, so take this with a grain of salt.

Mirror itself:
I think the mirror should go for implementing [FindFiles](https://dokan-dev.github.io/dokan-dotnet-doc/html/interface_dokan_net_1_1_i_dokan_operations.html#ad51643a31c941fc781181e8929d3b969) (only and leaving FindFilesWithPattern unimplemented to keep the implementation simple (like RegistryFS). I cannot see any advantage in implementing FindFilesWithPattern in the C#-mirror in the way it is currently proposed for this PR, but feel free to tell me what I am missing and/or tell me what benefit an implemented FindFilesWithPattern provides.

We can still add a comment inside the unimplemented FindFilesWithPattern hinting to DokanIsNameInExpression for those that want to implement their own FindFilesWithPattern.

That's just my opinion. Feel free to let me know why and where your opinion differs.

Liryna:
I agree with @Rondom on this.

TrabacchinLuigi:
Unfortunately i'm not a burocracy person ... I don't know what we can or not copy from that test...
I've removed the _ReactOS_ strings and i can say that i've revised it changing two test results because to me it looks like they where wrong, it is another test framework and another language ...

DokanNet/HelperMethods.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// <summary>
+        /// check whether Name matches Expression
+        /// Expression can contain "?"(any one character) and "*" (any string)
+        /// when IgnoreCase is true, do case insenstive matching
+        /// </summary>
+        /// <seealso cref="http://msdn.microsoft.com/en-us/library/ff546850(v=VS.85).aspx"/>
+        /// <param name="Expression">Matching pattern</param>
+        /// <param name="Name">Name to test</param>
+        /// <param name="IgnoreCase">Ignore case</param>
+        /// <returns></returns>
```
magol: - Have a capital letter at the beginning of sentences and point at the end.
- Explain more how this feature should be used, and when. Preferably with a reference to IDokanOperations.FindFilesWithPattern. To further understand how this should be used for the uninitiated, add some example code.
- Do not forget to write something in the `<return>` tag.

Rondom:
From a quick glance the PR looks good. Not a C#-expert. @C#-people: Is the HelperMethods-class necessary? Is there a better place or is it a good thing to introduce that class?

> Unfortunately i'm not a burocracy person ... I don't know what we can or not copy from that test...
I've removed the ReactOS strings and i can say that i've revised it changing two test results because to me it looks like they where wrong, it is another test framework and another language ...

The PR in its current form still has the ReactOS tests in it. (And no, changing all instances of "ReactOS" to something else does not change anything 😄 ).

Can anyone else give an opinion on this, to make sure I am not saying anything wrong? My understanding is that if we do not want to make the tests LGPL or create a separate library for the tests in question we cannot use their code (neither is worth it just for the sake of incorporating those tests in this codebase, IMO).

TrabacchinLuigi:
To me it is ok to move the method where ever you need, i actually disliked too having to create the helper class.
Let me know what to do also with the tests ...

Liryna:
Hi @TrabacchinLuigi ,

From my point of view,  DokanNetMirror.Tests test should be moved into DokanNet.Tests. Do not hesitate if you need help for this (currently not having the time but will be able later).

HelperMethods should be renamed to DokanHelper. I looked if could add it into another class but there are not other class that contained such methods.

About license @Rondom, I looked a little and yes we cannot copy/rewrite the test without breaking the license. We need to create our own test. We can look at their test value but we have to create our own (that will obviously look the same since we do the same test...).
Sorry @TrabacchinLuigi that his begin to be hard but license is serious 😢

TrabacchinLuigi:
No problems, I'll do it when i got a little spare time 😉

TrabacchinLuigi:
Is there something else i could/should do ?
lilbub

Liryna:
Such a Meow :cat: !

Thanks for the rename.

>From my point of view, DokanNetMirror.Tests test should be moved into DokanNet.Tests. Do not hesitate if you need help for this (currently not having the time but will be able later).

There is still this point :disappointed: to do

TrabacchinLuigi:
Give me some more instructions, because i thought that point was made with the rename... 😯

Liryna:
On the PR you created DokanNetMirror.Tests project, can DokanIsNameInExpression_should_match_as_expected be moved to DokanNet.Tests project ?
Like that we would have one project for all the test and DokanNetMirror.Tests will be removed.

What are the changes in DokanNet.net4.6.csproj , DokanNet.net4.0.csproj  ? (the diff on github show all file as changed)

https://github.com/dokan-dev/dokan-dotnet/pull/134/files#r100403581

Rondom:
@Liryna What about the license issue in the tests?

(For me it would be fine to go without or only very simple tests. Instead we should have something in winfstest or run one of the Microsoft tests like Runkarr or IFStest that test it in a black-box manner)

Liryna:
Yeah lets go without it 👍 
Will look to add Runkarr or IFStest one day 😢 

@TrabacchinLuigi just remove DokanNetMirror.Tests with DokanIsNameInExpression_should_match_as_expected we will do the test in another way later.

Rondom:
I might submit some followup-PR to Dokan Libary and (once-accepted also dokan-dotnet) to explain things a bit better. :-)

DokanNet/DokanHelper.cs
```diff
@@ -0,0 +1,160 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This method is mainly used in <see cref="IDokanOperations.FindFilesWithPattern"/> to filter a list of possible files.
+        /// <seealso cref="http://msdn.microsoft.com/en-us/library/ff546850(v=VS.85).aspx"/> 
+        /// </remarks>
+        /// <param name="Expression">The matching pattern. Can contain: ?, *, &lt;, &quot;, &gt;.</param>
+        /// <param name="Name">The string that will be tested.</param>
+        /// <param name="IgnoreCase">When set to true a case insensitive match will be performed.</param>
+        /// <returns>Returns true if Expression match Name, false otherwise.</returns>
+        /// <example>"F0_&lt;&quot;*" match "f0_001.txt"</example>
```
magol: Doxygen do ignore the `<example>` tag, see https://www.stack.nl/~dimitri/doxygen/manual/xmlcmds.html. Move the example to some other tag.

DokanNet/DokanHelper.cs
```diff
@@ -0,0 +1,160 @@
...
+        public static Boolean DokanIsNameInExpression(String Expression, String Name, Boolean IgnoreCase)
+        {
```
magol: Add some guard check of the string arguments?

DokanNet/DokanNet.net4.0.csproj
```diff
@@ -1,132 +1,133 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
magol: Why is all lines in this file changed?

DokanNet/DokanNet.net4.6.csproj
```diff
@@ -1,133 +1,134 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
magol: Why is all lines in this file changed?

DokanNet.Tests/DokanOperationsFixture.cs
```diff
@@ -343,7 +343,29 @@ public NtStatus WriteFile(string fileName, byte[] buffer, out int bytesWritten,
                 => TryExecute(fileName, buffer, out bytesWritten, offset, info, (string f, byte[] b, out int w, long o, DokanFileInfo i) => Target.WriteFile(f, b, out w, o, i), nameof(WriteFile));
         }
 
-        public const string MOUNT_POINT = "Z:";
+        private static String _mount_point;
+
+        public static String MOUNT_POINT
```
magol: use `string` instead of `String` for `MOUNT_POINT` and `_mount_point`

DokanNet/DokanHelper.cs
```diff
@@ -0,0 +1,160 @@
...
+using System.Text;
+
+namespace DokanNet
+{
+    public class DokanHelper
+    {
+        /// <summary>
+        /// Matches zero or more characters until encountering and matching the final . in the name.
+        /// </summary>
+        const Char DOS_STAR = '<';
```
magol: Use `char` instead of `Char` for all constants.

DokanNet/DokanHelper.cs
```diff
@@ -0,0 +1,160 @@
...
+        /// <remarks>
+        /// This method is mainly used in <see cref="IDokanOperations.FindFilesWithPattern"/> to filter a list of possible files.
+        /// <seealso cref="http://msdn.microsoft.com/en-us/library/ff546850(v=VS.85).aspx"/> 
+        /// </remarks>
+        /// <param name="Expression">The matching pattern. Can contain: ?, *, &lt;, &quot;, &gt;.</param>
+        /// <param name="Name">The string that will be tested.</param>
+        /// <param name="IgnoreCase">When set to true a case insensitive match will be performed.</param>
+        /// <returns>Returns true if Expression match Name, false otherwise.</returns>
+        /// <example>"F0_&lt;&quot;*" match "f0_001.txt"</example>
+        public static Boolean DokanIsNameInExpression(String Expression, String Name, Boolean IgnoreCase)
```
magol: Use `bool `instead of `Boolean` and `string `instead of `String`.

DokanNet/DokanHelper.cs
```diff
@@ -0,0 +1,160 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
```
magol: Following can be removed:
```C#
using System;
using System.Collections.Generic;
using System.Text;

```

DokanNet/DokanHelper.cs
```diff
@@ -0,0 +1,160 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace DokanNet
+{
+    public class DokanHelper
```
magol: Make the class static

DokanNet/DokanHelper.cs
```diff
@@ -0,0 +1,160 @@
...
+        /// Matches zero or more characters.
+        /// </summary>
+        const Char ASTERISK = '*';
+
+        /// <summary>
+        /// Matches a single character.
+        /// </summary>
+        const Char QUESTION_MARK = '?';
+
+        private static Char[] CharsThatMatchEmptyStringsAtEnd = new Char[] { DOS_DOT, DOS_STAR, ASTERISK };
```
magol: Change to 
````C#
private static readonly char[] CharsThatMatchEmptyStringsAtEnd = { DOS_DOT, DOS_STAR, ASTERISK };

File: DokanNet/DokanHelper.cs [hunk 12]
```diff
@@ -0,0 +1,160 @@
...
+                    else
+                    {
+                        return false;
+                    }
+                }
+            }
+
+            var nextExpressionChars = Expression.Substring(ei);
+            var AreNextExpressionCharsAllNullMatchers = Expression.Any() && !String.IsNullOrEmpty(nextExpressionChars) && nextExpressionChars.All(x => CharsThatMatchEmptyStringsAtEnd.Contains(x));
+            var IsNameCurrentCharTheLast = ni == Name.Length;
```
magol: lower case on `areNextExpressionCharsAllNullMatchers` and `isNameCurrentCharTheLast`.

DokanNet/DokanHelper.cs
magol: Is it possible to lower the cyclomatic complexity down from 34?

TrabacchinLuigi:
Lol, my bad habit of using actual types instead of primitives :D I'll do as much as I can, not sure about the cyclomatic complexity and why lines changes on csprojs, maybe vs 2017 did that

magol:
@TrabacchinLuigi It is easy to get bad habits :-)

The cyclomatic complexity is not that important. 

I do not understand what VS2017. Maybe you have do restore the project files and make the changes manually? It is stil some bugs in VS2017 RC (I use it myself)

TrabacchinLuigi:
VS2017 is Visual Studio 2017 i'm testing it a bit.
Hehe i kinda like using the types from System, because they get colored as types (green) in visual studio , makes more sense to me 😄 my colleague hate that :trollface: 
it's like tabs vs spaces, endless war 😆

PalmaNicola:
Luigi, I beg you, please stop doing it.

TrabacchinLuigi:
I have no idea if this will fix the fact that git consider all the lines in the csprojs changed, i tried to compare them also with notepad++ but can't see any change ...

Liryna:
Try `git checkout 80278ee221144df8ba2b0f8be6807bf3d170981c DokanNet/DokanNet.net4.0.csproj DokanNet/DokanNet.net4.6.csproj`
It will reset the csproj to the current master

TrabacchinLuigi:
did it, re added the missing line, same result ...

TrabacchinLuigi:
can we flat this to a single commit somehow ?

Liryna:
Yes, you can rebase all your commit into one with git rebase;
https://www.atlassian.com/git/tutorials/rewriting-history

TrabacchinLuigi:
i don't have tortoise, just git for windows, and github addin for visual studio.
The guide says do not do that if you published the changesets ... did anyone did that before ?
I'm afraid doing that will fuck up this discussion ... and the pull request ...

Liryna:
Sorry for the delay @TrabacchinLuigi 
No problem, I have merged your changes into two commits https://github.com/dokan-dev/dokan-dotnet/commit/b26a339306f4714ec29917b8021175f5b5be0fad https://github.com/dokan-dev/dokan-dotnet/commit/065711e82b700b1958b11b17a37f14e4b4235e2b

Thanks again for you help and contribution ! 👍 (loved this cat)

Issue #137, opened 2017-03-02
Visual Studio 2017


magol:
On March 7 is Visual Studio 2017 [launched](https://blogs.msdn.microsoft.com/visualstudio/2017/02/09/visual-studio-2017-launch-event-and-20th-anniversary/) . One of the big news is that csproj file will be greatly simplified and supports multi-targets.

Additionally, Microsoft now recommend that we should use [.NET standard](https://docs.microsoft.com/en-us/dotnet/articles/standard/library) for packages. In this case, we should replace the .NET Framework 4.6 with .NET 1.3 Standard 

Today we have two separate projects, one for .NET 4.0 and .NET 4.6. I recommend that we replace these with a single project in VS2017-format that multi-target to `net4.0` and `netstandard1.3` (the same as .NET 4.6).

## Open questions
* When do AppVeyor support Visual Studio 2017?
* Is it possible to open the new csproj-file in older Visual Studio (probably not). 
  * If not possible to open in older Visual Studio, is it a problem?
     * If it is a problem, how long should we wait for all devs to update to Visual Studio 2017?
* Should we maybe have a separate csproj file in the old format to make it possible for users of old Visual Studio to compile? It is enough that there is a project file only for .NET 4.6 (or .NET 4.0). The downside is that the environment becomes more complicated when we have to make sure to keep this project file up to-date when we do not use it.

Rondom:
> When do AppVeyor support Visual Studio 2017?
They do have a beta and I think I have read that they plan to make it official soonly after VS2017 is released.

Liryna:
👍 This is a good idea, it should be made when VS2017 is released. I guess they will make a community version for people to get it in any case.
(hope that VS 2017 will be more stable that VS 2015)

Liryna:
What about nuget, is it already capable to install properly  .NET standard library ?

magol:
@Liryna Yes, VS2017 do properly install .NET standard libraries from NuGet.
But it is still some bugs in VS2017 :-(

TrabacchinLuigi:
Net standard and xprojs have been a pain, but now it looks like they are not going to change much more, new csprojs are quite easy to configure and I think a retro compatibility for vs2015 patch will enable to build those projects soon... Also vs2015 and 2017 live side by side with no issues, we will install vs2017 as soon as it is released.
Kinda worth to target more projects...

Rondom:
[...]

PR #138, opened 2017-03-02
No merge - Migrate to VS2017

magol:
#137


Rondom:
Maybe you already know it, but In case you need to troubleshoot the AppVeyor-Build, [you can enable RDP](https://github.com/dokan-dev/dokany/blob/6cfc9c143fdc713327311feaff075311b5f9584b/appveyor.yml#L27-L33).

magol:
@Liryna 
I'm getting happy with it now. I have not yet been able to test netstandard1.3, MStest lacks functionality for DataRow when running under .NET Core 1.X. Perhaps we can fix this in a separate PR.

Can you please review what I have done. When everything is ready for merge, I'll clean up all the commits.

Liryna:
@magol thank you for all the changes !
I am just wondering since I do not have vs 2017 for now, what happened to all the csprojet, assembly info and such ? Looks like everything has been removed and see nothing new that replace them ? What about nuget package nuspec also removed ?

As my understanding, Dotnet core 1.x do not provide everything by your comments. Do you think we should request the missing features ?

Otherwise this seems to be alright for me ! I will just try to get vs 2017 soon.

magol:
The best change in vs2017 is that csproj has become really simplified. All redundant and unnecessary information is deleted, and the management of the nugget is now native in the csproj file.
Even that was previously found in AssemblyInfo is now in the csproj file. 

As i have understod it, most of the missing parts of .NET core will come back in version 2.0. In that case, we may add the missing tests when it is launched in the fall.

Liryna:
Ok that makes clearer 👍 Is there already a date for version 2.0 ?

magol:
@Liryna From [the roadmap](https://github.com/dotnet/core/blob/master/roadmap.md)

| Milestone	| Release Date |
|----------------|---------------|
| .NET Core 2.0 Preview |	Q2 2017 |
| .NET Standard 2.0 Preview |	Q2 2017 |
| .NET Core 2.0 | Q3 2017 |
| .NET Standard 2.0| Q3 2017|

Liryna:
Hi @magol ,

I just tried to build the solution with vs 2017 and have an error:
```
AppData\Local\Temp\.NETStandard,Version=v1.3.AssemblyAttributes.cs(4,20,4,26): error CS0400: The type or namespace name 'System' could not be found in the global namespace (are you missing an assembly reference?)
```
Am I missing something ?

Otherwise, how can I build for framework 4 ?

magol:
I've got too this error a few times. I think VS2017 is a bit buggy when it comes to this.
Start the command prompt, go to the project folder and run the following command to fix it:
`dotnet restore`

Since I use `<TargetFrameworks> netstandard1.3; net4.6; net4.0 </ TargetFrameworks>`, VS compiles all this targets simultaneously when compiling. To select which version of Mirror you want to run, select in the list as shown below:
image

If you only want to build for only one target, change to `<TargetFramework>net4.0</TargetFramework>` in the project file (note that I removed the 's' to make it singular).

Liryna:
Build work perfectly for me ! That’s really a great work 👍 
(it is even the first time I am impressed by vs2017 😮 )

Does `dotnet restore`need to be run for each project clone on it is system first launch ?

magol:
Great :-D

I don't know if `dotnet restore` have to be run each time you clone it to a new system. But my experience is that you have to for netstandard applications. The command is used to restore all NuGet packages, strange that VS can not do it for you :-(

Liryna:
@magol Thanks a lot !
Do you think vs2017 / PR is ready to be merged ?

magol:
Yes, I think it's ready to merge :-)

I do not know if you have already realized this, but I would still like to give you some tips and other notes:

- You can now easily edit the project file from within Visual Studio without having to unload the project. Just right-click on the project and choose `Edit DokanNet.csproj`. This must be done when editing version numbers and information about NuGet.

- Speaking of NuGet, A nupkg file is now automatically created every time you compile and is placed in bin\Release. This is due to `<GeneratePackageOnBuild>True</GeneratePackageOnBuild>`

- I have add some tags to the NuGet. That was missing before, but I think it is important to make it easier to find on nuget.org. Some that you think are missing (or are unnecessary)? It's hard to find good search words.
`<PackageTags>dokan file files disk directory storage filesystem io filestore FAT NTFS FUSE</PackageTags>`

Liryna:
Alright 👍 let’s do this !

So the version cannot still be updated from interface directly ? Or even taken from a define/macro of the code ?

Create nuget package by default is alright for me and the tags list you made is also correct 👍 

Thanks a lot again @magol it is really a great contribution !

magol:
@Liryna 
Of course, you can also update the version number from the Properties :-)

But I'm considering maybe improving the version number management so that we only needed to updating it in one place. Now, the sample programs do not get a version number and I like to fix it.

My idea is to have a `version.props` that all csproj files imports. In that way, it is enough to change only that file, and all assemblies get the same version number. Maybe even make a script that updates the Doxygen file?

Liryna:
Yes 👍 having the version only in one place would be the best

Rondom:
No need to update the Doxyfile, you can set the number on command line. Maybe one can even invoke doxygen from msbuild?
Btw I have started to do something similar for the main Dokan project. Not really finished, but I can push my current progress. 
There the challenge was that there are many places that hard code the version number in one way or the other. Also making cmake (used for Cygwin and MinGW FUSE build) parse the MSBuild-XML file is not really a viable approach so I opted for a simple file with just the version that can be read easily by MSBuild, CMake, PowerShell, you name it.

magol:
You right, to set the doxygen version from the command line is much beter :-)
To invoke doxygen from msbuild was clever. In that way, you can review the documentation before push it to GitHub. But then you have to download doxygen.exe to the developing machine, manual or automatic. I don't know if that is possible (or easy) to do it automatic from MsBuild. That is something to investigate.

It is valuable to have the same approaches for both Dokany and Dokan-dotnet. I'm very interested in how you've done it (even if it's not ready), where do I find it.

Having a simple text file with the version number is probably the best way as you say. Then you are more free to handle it as you want.

Rondom:
I am by no means a MSBuild expert and I have no intention of becoming one. So, take the following with a pinch of salt. Can you make the Doxygen-build optional in some way? Like a separate build target or so. I mean it's not like the Doxygen invocation will be different among .NET versions, Debug/Release config etc. I do not use Visual Studio, so I may not really know what I am talking about...
About dependencies: I also considered writing some wizard-like script, that guides the developer in installing all dependencies of Dokan (Dokan has quite a few if you want to build everything), which makes dependencies like Doxygen more bearable.

This is my current progress: https://github.com/Rondom/dokany/tree/no_hardcode_version
I decided on using Property-Expressions for reading the file, because I could not get a build-step to set a preprocessor-define (You need to include your file at the end of the project to hook into the vcxproj build-process to do something BeforeBuild. At this point the variables in `PreprocessorDefinitions` are already evaluated it seems). As I said, I have never done anything with MSBuild so the above may be wrong.

About the current state of my branch here:
* MSBuild uses the version for resource-files and INF-files
* cmake uses the version for dokan_fuse resource-files
* everything else (Doxygen, Installer, AppVeyor, some consistency checks) is not done yet (see commit message)
* I rebased recently without having a machine to test the building, so the current state could be broken. I do not have the time to test and polish it atm, Sorry.

Issue #144, opened 2017-05-02
Timeout wrong when Mounting


TimJRodgers:
In the method Mount(), it takes a TimeSpan for the timeout, which it then assigns to the DokanOptions struct. However, it uses the wrong property from the TimeSpan. The line;
Timeout = (uint) timeout.Milliseconds

Should be;
Timeout = (uint) timeout.TotalMilliseconds

If I create the time span as a whole number of seconds such as TimeSpan.FromSeconds(20), then the time out will be 0 rather than 20000. (And consequently the highest time out I can set is 999ms.)

Thanks,

Tim.

magol:
@TimJRodgers  
Wow, great that you found this. I've made a fix on what comes with the next version.

TimJRodgers:
That's great, thank you!

Liryna:
Thanks a lot @magol for the quick fix and @TimJRodgers for the report ! 🏆

PR #145, opened 2017-05-03
Corrects the use of timeout in mount

magol:
See #144


Liryna:
Perfect @magol ! Thanks a lot ! ❤️

Issue #136, opened 2017-02-09
SetTime while stream is open


TrabacchinLuigi:
I found out that some software update the last write time while they still have the stream opened, this cause it to fail because of the simplified API the framework exposes.
The framework open a stream to obtain a file handle and use it to pinvoke windows api, obviously it can't if there is another stream already opened ... 
I've already implemented a solution for that. I've copied what the framework does, Seems that it works... I will update you with some code, and maybe some failure examples...
You can reproduce it using Revit and the mirror.cs saving some workshared projects it will fail.

TrabacchinLuigi:
Some documentation:
[System.IO.File.SetCreationTimeUtc](https://referencesource.microsoft.com/#mscorlib/system/io/file.cs,63bd669b43be5f17,references)
From this we can understand that this data is not in the file, but in the filesystem, so if we want to mirror it using .net we need to use the file handle and windows API

Liryna:
Hi @TrabacchinLuigi ,

This is true, it is how windows filesystem work. I don’t think we can improve the situation.
FileStream give access to the handle so it is "easy" to call the win32 API like you said.

Otherwise it is still possible to not open the file with the proposed to share access and override it by something that would let the SetCreationTimeUtc.

TrabacchinLuigi:
My proposal is to fill a bug to the .net team and create a method in SafeFileHandle to set the time, meanwhile i'm creating a nuget packet that have an extension method to do that... what you think ?

TrabacchinLuigi:
I can't find where to fill this proposal/bugfix to the .net team 🙄

magol:
I think https://github.com/dotnet/corefx is the right place for proposal to  .Net. But i can be wrongly.

TrabacchinLuigi:
I gave it a shot, I'll be happy if you guys could join the [discussion](https://github.com/dotnet/corefx/issues/16187) and say what you think.

PR #151, opened 2017-08-01
Fix Mirror.cs implementation of SetTime to work on open files.

Deathspike:
This solves https://github.com/dokan-dev/dokan-dotnet/issues/136

### DIFF
--- a/sample/DokanNetMirror/Mirror.cs
+++ b/sample/DokanNetMirror/Mirror.cs
@@ -151,7 +151,7 @@ namespace DokanNetMirror
                         if (pathExists)
                         {
                             if (readWriteAttributes || pathIsDirectory)
-                            // check if driver only wants to read attributes, security info, or open directory
+                                // check if driver only wants to read attributes, security info, or open directory
                             {
                                 if (pathIsDirectory && (access & FileAccess.Delete) == FileAccess.Delete
                                     && (access & FileAccess.Synchronize) != FileAccess.Synchronize)
@@ -193,7 +193,7 @@ namespace DokanNetMirror
                         readAccess ? System.IO.FileAccess.Read : System.IO.FileAccess.ReadWrite, share, 4096, options);
 
                     if (pathExists && (mode == FileMode.OpenOrCreate
-                        || mode == FileMode.Create))
+                                       || mode == FileMode.Create))
                         result = DokanResult.AlreadyExists;
 
                     if (mode == FileMode.CreateNew || mode == FileMode.Create) //Files are always created as Archive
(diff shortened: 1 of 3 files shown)


sample/DokanNetMirror/Win32.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Runtime.InteropServices;
+using Microsoft.Win32.SafeHandles;
+
+namespace DokanNetMirror
+{
+    public class Win32
```
magol: - Please add XML comments to the class and method. See [NativeMethods.cs](https://github.com/dokan-dev/dokan-dotnet/blob/master/DokanNet/Native/NativeMethods.cs) for inspiration.
- Please rename the class to `NativeMethods`, see [CA1060: Move P/Invokes to NativeMethods class](https://msdn.microsoft.com/en-us/library/ms182161(v=vs.110).aspx)

sample/DokanNetMirror/Mirror.cs
```diff
@@ -380,7 +380,18 @@ public NtStatus SetFileAttributes(string fileName, FileAttributes attributes, Do
         {
             try
             {
+                var stream = info.Context as FileStream;
+                if (stream != null)
+                {
+                    var ct = creationTime?.ToFileTime() ?? 0;
+                    var lat = lastAccessTime?.ToFileTime() ?? 0;
+                    var lwt = lastWriteTime?.ToFileTime() ?? 0;
+                    if (Win32.SetFileTime(stream.SafeFileHandle, ref ct, ref lat, ref lwt)) return NtStatus.Success;
```
magol: Please move the return to a new line.

You can have different tastes, but I would have preferred if it looked like this:
```C#
if (!Win32.SetFileTime(stream.SafeFileHandle, ref ct, ref lat, ref lwt))
{
    throw Marshal.GetExceptionForHR(Marshal.GetLastWin32Error());
}
return NtStatus.Success;

```

sample/DokanNetMirror/Mirror.cs
```diff
@@ -151,7 +151,7 @@ private string GetPath(string fileName)
                         if (pathExists)
                         {
                             if (readWriteAttributes || pathIsDirectory)
-                            // check if driver only wants to read attributes, security info, or open directory
+                                // check if driver only wants to read attributes, security info, or open directory
```
magol: Maybe move the comment to above the `if`

sample/DokanNetMirror/Mirror.cs
```diff
@@ -380,7 +380,18 @@ public NtStatus SetFileAttributes(string fileName, FileAttributes attributes, Do
         {
             try
             {
+                var stream = info.Context as FileStream;
+                if (stream != null)
```
magol:
```C#
var stream = info.Context as FileStream;
if (stream != null)
```
can be replaced width this in [C#7](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#pattern-matching)
```C#
if (info.Context is FileStream stream)
```

Liryna:
@Deathspike thank you for the PR it is very welcome !
@magol has reviewed your PR with some pending questions. Can you look at them ?

Liryna:
@magol do you think we can merge this an fix regarding your review ?

magol:
@Liryna  Yes, I think we can merge this anyway :-)

Liryna:
@magol Can you take care of it and apply changes of your reviews ❤️ ?

Liryna:
Thank you @magol for the merge and fix !!! ❤️

PR #156, opened 2017-10-12
 Fix comments for PR #151 and update icon

magol:


Liryna:
Ok for me ! and good idea to change the ico also !

PR #161, opened 2017-11-29
Adapt to API changes for dokany 1.1.0

0x4a616e:
I updated the native call to `DokanMapKernelToUserCreateFileFlags` according to the changes in https://github.com/dokan-dev/dokany/commit/0da77870161c101d3133fd76938951c200f12868

### DIFF
--- a/DokanNet.Tests/DokanOperationsFixture.cs
+++ b/DokanNet.Tests/DokanOperationsFixture.cs
@@ -388,6 +388,8 @@ namespace DokanNet.Tests
 
         private long pendingFiles;
 
+        public static bool HasPendingFiles => Instance?.pendingFiles > 0;
+
         internal static IDokanOperations Operations => proxy;
 
         internal static DokanOperationsFixture Instance { get; private set; }
--- a/DokanNet.Tests/DokanOperationsFixture.cs
+++ b/DokanNet.Tests/DokanOperationsFixture.cs
@@ -984,7 +984,7 @@ namespace DokanNet.Tests
         internal void ExpectOpenDirectoryWithoutCleanup(string path, FileAccess access = FileAccess.Synchronize, FileShare share = FileShare.None, FileAttributes attributes = EmptyFileAttributes)
         {
             operations
-                .Setup(d => d.CreateFile(path, access, share, FileMode.Open, ReadFileOptions, attributes, It.Is<DokanFileInfo>(i => i.IsDirectory)))
+                .Setup(d => d.CreateFile(path, FileAccessUtils.MapSpecificToGenericAccess(access), share, FileMode.Open, ReadFileOptions, attributes, It.Is<DokanFileInfo>(i => i.IsDirectory)))
                 .Returns(DokanResult.Success)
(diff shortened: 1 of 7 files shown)


0x4a616e:
Unfortunately i was having a hard time to get the tests running due to some nugget issues, so I'm now adding in some belated test fixes 😬 The last commit seemed to have fixed most file access rights related issues. However, I was having some more test failures on my machine. As an example, in `DriveInfoTests.GetDriveFormat_CallsApiCorrectly()` the `CreateFile` method gets called with `FileShare.ReadWrite` instead of `FileShare.None`. I haven't changed this, as it is probably platform specific.
I can't tell why all tests are failing in the AppVeyor build, could this have something to to with the used dokany version?

Rondom:
Try changing the Dokan version to 1.1.0 in appveyor.yml. I don't know if that is the only problem, though.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Liryna:
Hi @0x4a616e ,

Thank you doing to API update !

@Rondom is right, appveyor dokan download link need to be updated here: 
https://github.com/dokan-dev/dokan-dotnet/blob/master/appveyor.yml#L26

Regarding the error on winfstest, the FS does not seem to be mounting (no pid to dismount at the end).
https://ci.appveyor.com/project/Liryna/dokan-dotnet/build/1.0.0-369/job/fbykt92qqcac3ee5#L457

This probably comes from appveyor images update since previous commit since nothing changes.

Otherwise yes...the mocks are kind hard to make work on every system, we should only focus for appveyor and update it when appveyor change behavior.
A simple new software on the system that wants to access the FS during the test, and it will break the expected test access.

0x4a616e:
thx for the tip, this is already looking _much_ better. It looks like appveyor is having the same issue as i described above, where FileShare is `FileShare.ReadWrite` instead of `FileShare.None`. I'll also give the winfstests a try and see if i can contribute anything there.

0x4a616e:
Hmm, I guess fixing this is probably a bit more difficult than I anticipated because all winfstests pass on my machine without any issue. It is a bit suspicious though that it's once again an issue related to shared file access...

Liryna:
@0x4a616e I had the same issue with winfstest, run ok on my machine and not on appveyor. Probably a race call issue from what I remember.

You can let it as it is.
But If your a motivated to look for the dotnet failure on release it would be amazing to get appveyor on green again :+1:

0x4a616e:
The issue seems to be a race condition where the test initialise method is called before the mounting is completed. This is probably the reason why it only hits the first test case. The mounting process still has a pending `CloseFile` operation when the test initialise method resets the mocks.
I now added an additional delay loop to the `Mounter` to wait for all pending files to be closed before continuing to run the actual tests. Did the trick for me.

Liryna:
@0x4a616e this is perfect!
Thanks a lot for updating the API and fixing the test :+1: It is a great job :trophy: and very appreciate !

I will do the release hopefully still today otherwise tomorrow on nuget.

Issue #174, opened 2018-04-26
Does the handle returned by OpenRequestorToken need to be closed?


kyanha:
Does the handle returned by OpenRequestorToken need to be closed?  If so, there's a resource leak in DokanFileInfo.GetRequestor(), which obtains the handle, then uses it to construct a System.Security.Principal.WindowsIdentity.  Per the reference source, WindowsIdentity calls Win32Native.DuplicateHandle on it after it verifies that it's valid.  The original goes out of scope without being closed.

kyanha:
(Since handle values of both 0 and -1 seem to be invalid to WindowsIdentity, I'd suggest a subclass of SafeHandleZeroOrMinusOneIsInvalid would be appropriate.  And since it doesn't need to be used outside of that function, might I suggest SafeFileHandle?  Otherwise, I can whip up a quick SafeTokenHandle class... but that seems a bit overkill for just one spot.)

Liryna:
Hi @kyanha ,

Thank you for the report, you are right ! The handle needs to be closed and I agree on the solution you suggest. Do you think you could make a PR with the changes ? 🚀

kyanha:
As an aside, do you know anyone who has the proper P/Invoke declarations for the functions in authz.dll?  That's Microsoft's user-mode component that provides access to a vetted and correct ACL evaluation implementation.

kyanha:
(I just looked at the Appveyor test results.  This patch doesn't touch any code that it complains about?)

kyanha:
Pull request #176 rebases the 3 prior commits I made into one unified patch.  (I have closed pull request #175.)

PR #175, opened 2018-05-09
Fixed leak of the token handle. #174

kyanha:
It turns out that the WindowsIdentity constructor doesn't take
ownership of the handle that it's passed.  As a result, the handle
returned directly from DokanOpenRequestorToken never gets closed.
Constructing a subclass of SafeHandle around it ensures that it
will be closed eventually, if it's used in a using() block.  (We
do this because there are some situations where a finally{} block
won't get executed, though we also use the finally mechanism to
attempt to dispose of it as soon as we're actually done with it.)
Because handle values of both 0 and -1 are invalid as token
handles (per the Reference Source for WindowsIdentity), we need
a subclass of SafeHandleZeroOrMinusOneIsInvalid; it turns out that
SafeFileHandle is the only concrete class that derives from it.
We could implement a SafeTokenHandle class which derives from the
same base class, but that seems like a lot of work for only one
location where it's used.  So, we construct it as a SafeFileHandle
and expect that it won't be a problem.  (If Microsoft ever
revisits the SafeFileHandle code to add in a check that the
handle it's constructed with is actually a file handle, we will
need to implement SafeTokenHandle.)

### DIFF
--- a/DokanNet/DokanFileInfo.cs
+++ b/DokanNet/DokanFileInfo.cs
@@ -150,7 +150,10 @@ namespace DokanNet
             }
             finally
             {
-                sfh.Dispose();
+                if (sfh != null)
+                {
+                    sfh.Dispose();
+                }
             }
         }
 
--- a/DokanNet/DokanFileInfo.cs
+++ b/DokanNet/DokanFileInfo.cs
@@ -136,9 +136,10 @@ namespace DokanNet
         /// -or- <c>null</c> if the operation was not successful.</returns>
         public WindowsIdentity GetRequestor()
         {
+            SafeFileHandle sfh = null;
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @kyanha ,

Thank you for the report and also the contribution ! Very appreciate 🏆 
Can you just rebase all the commit into one ?

Do not worry about the unit test, I need to fix them 😃

Liryna:
@kyanha I have fixed the test, you can rebase your commit with it to rerun the test.
https://ci.appveyor.com/project/Liryna/dokan-dotnet/build/1.1.0-386

kyanha:
Replaced with PR #176.

Issue #177, opened 2018-05-13
DokanFileInfo leaks GCHandle if DFI.Context not set to null in CloseFile


kyanha:
DokanFileInfo allocates a GCHandle when DFI.Context is set, but it only ever frees it if DFI.Context is set to null thereafter.  Neither CleanupProxy nor CloseFileProxy set it to null before ending its referenceability.  This causes a GCHandle to be leaked, and there are only a finite number of these that can be created before the runtime fails.

I would like to make a test in CloseFileProxy to see if the object pointed to by DFI.Context is IDisposable, and if it is to call Dispose() on it, before setting DFI.Context to null in all cases.  This will permit the object to clear itself from anything it might add to (a list or dictionary of handles that are open to a given file for purposes of determining when DeleteFile's processing should actually go ahead, for example), before causing the GCHandle to be freed.

(Because access to the file after CleanupProxy is still possible in the case of DokanFileInfo.PagingIo, I don't think that CleanupProxy is the appropriate place to do this.)

kyanha:
Pull Request #179, putting it in the finally {} block.

Liryna:
Hi @kyanha ,

I understand your point of view. I would agree on your change if the logic would be that the Context is owned by Dokan all the time but here it is the user fs that set the Context. In another word, Dokan should not be the one managing the life of it.

I know that there is FS that use Context only as information and keep the real data/context is their own system to release/handle it later after Close or event at the unmount. Dispose would not be the behavior they want.

Close is always called, if the user FS do not Dispose it, it is an issue they need to handle.

kyanha:
If the object in the Context isn't IDisposable, it won't have its Dispose method called.  In the case you cite, it sounds like they use the content of Context as a key into their own storage system, and the type of that key probably doesn't implement IDisposable.  If they implement the IDisposable pattern correctly, they definitely want it to be called sometime (perhaps as late as when the finalizer runs).  I just think that ensuring the heap isn't full of Context objects that relate to dead file handles is something the library could help the programmer do.  (After all, if they don't want the Dispose method called on their context object, they don't have to implement IDisposable on it.)

That said, I'll accept your point.

But that's all secondary to the main problem of leaking GCHandles, which (unlike Context) really is owned by Dokan all the time.  Will you accept the patch if I get rid of the call to Dispose?

Liryna:
[...]

PR #179, opened 2018-05-13
Issue #177: DokanFileInfo.Context leaks GCHandle if not set null

kyanha:
DokanFileInfo.Context allocates a GCHandle if it's set to anything
that isn't null.  However, that GCHandle is only ever freed if it
is set to null thereafter.  This patch ensures that the GCHandle
is proprly freed before the DokanFileInfo.Context goes out of
scope.  It also checks to see if the object assigned to it can be
cast to IDisposable, and if it can it calls its .Dispose() method.
This allows for context objects to remove references to
themselves, from (e.g.) lists or dictionaries of open files used
to figure out when the file they refer to should actually be
deleted (which should only happen once all handles to the file are
closed).

### DIFF
--- a/DokanNet/DokanOperationProxy.cs
+++ b/DokanNet/DokanOperationProxy.cs
@@ -355,6 +355,11 @@ namespace DokanNet
             {
                 logger.Error("CloseFileProxy : {0} Throw : {1}", rawFileName, ex.Message);
             }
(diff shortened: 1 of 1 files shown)


kyanha:
Superseded by #180, which was merged.

PR #180, opened 2018-05-13
Issue #177: DokanFileInfo.Context leaks GCHandle if not set null.

kyanha:
DokanFileInfo.Context allocates a GCHandle if it's set to anything
that isn't null. However, that GCHandle is only ever freed if it
is set to null thereafter. This patch ensures that the GCHandle
is proprly freed before the DokanFileInfo goes out of scope,
without calling anything on the object the Context points to.


Liryna:
Thank you @kyanha for the quick PR and change !

Issue #181, opened 2018-05-13
DokanFileInfo doesn't expose the Object ID members of BY_HANDLE_FILE_INFORMATION


kyanha:
DokanFileInfo doesn't expose the Object ID members of BY_HANDLE_FILE_INFORMATION, which would be 8 bytes long (or an Int64 -- it can't be a UInt64 because UInt64 is not CLS-compliant).

Exposure of these members seems to be required to legitimately claim FileSystemFeatures.SupportsObjectIDs.

If the filesystem claims that it supports Object IDs, they are supposed to stay the same, regardless of their name or the volume they reside on.  Currently, the BY_HANDLE_FILE_INFORMATION.nFileIndexHigh (high 32 bits) are set to 0, and .nFileIndexLow is set to the HashCode of the file name.  This happens in GetFileInformationProxy, after the filesystem implementation of GetFileInformation returns to it.

I propose exposing the Object ID from DokanFileInfo, and if the Int64 is left untouched after CreateFile, to fall back to the current behavior of setting the low dword to the HashCode off the file name in GetFileInformationProxy.  (If the Int64 is set, then set the corresponding fields in the B_H_F_I structure to the high and low part of that value.)

As this is merely an extension of a current structure, I don't think it's a breaking change.

Does anyone know why the Object ID wasn't exposed by Dokan.Net before now?

kyanha:
Exposure in PR #182.  I'm not sure it's complete, though, given the lack of ObjectID (or nFileIndexHigh/nFileIndexLow) anywhere near CreateFile; this would be necessary both to open by ObjectID and to set the ObjectID on a file that's being copied from another volume.

PR #182, opened 2018-05-13
PR #181: expose ObjectID in FileInformation.cs

kyanha:
Currently (before this patch), the two fields of BY_HANDLE_FILE_INFORMATION called nFileIndexHigh
and nFileIndexLow are set according to the following rule: nFileIndexHigh is always 0, and nFileIndexLow is either the HashCode of the filename or 0 (if GetHashCode() fails).  This does not match the required behavior of ObjectIDs, which mandates that object IDs must stay the same for the same file, even in the face of moving or renaming it -- even across volumes.  As a result, exposing it is required for anything claiming FileSystemFeatures.SupportsObjectIDs.

There is most likely a lot more work to be done (finding a way to specify an object ID when a file is created, for example), but this is the barest bit of functionality that will allow user filesystems to start working toward ObjectID support.


kyanha:
Closing because it'll take a while for the kernel to do anything with the ObjectID (I do not feel anywhere near confident enough to muck around in kernel code, and I can't afford to hire anyone else to do it for me, so it most likely won't go anywhere for a while).

Issue #183, opened 2018-05-20
create instance of DokanFileInfo for testing


ghosttie:
In order to avoid blue screens while my file system is in an early stage, I'm implementing the IDokanOperations functions one at a time and testing them just in .NET, but to test some of the functionality I need to be able to create an instance of DokanFileInfo, which isn't currently possible.

Is there any chance you could add a way to create instances of DokanFileInfo for testing?

Liryna:
Hi @ghosttie ,

Your request is to expose DokanFileInfo outside the dll ?
The simplest would be to copy the class temporarily in your code.

Otherwise than this, I don't think this class has a purpose other than in IDokanOperations .

ghosttie:
I'd like to test IDokanOperations using .NET code without using the driver, but I can't do that because there isn't a constructor on DokanFileInfo.

If you wanted to support this type of testing you could use the impostor pattern: add an IDokanFileInfo interface which DokanFileInfo implements.  IDokanOperations would be changed to expect an IDokanFileInfo instead of a DokanFileInfo, and then a test class that implements IDokanOperations could be used for testing instead of a real DokanFileInfo.

Liryna:
OK I see 👍  I will let this open to see if others would like to see it happen.

kyanha:
I created a testing framework of this kind by basically creating wrappers around 

There's a way to do it using System.Runtime.Serialization and System.Reflection.  About the only thing you can't do with this is call OpenRequestor():

```cs
using DokanNet;
using System.Diagnostics;
using System.Reflection;
using static System.Reflection.BindingFlags;
using System.Runtime.Serialization;

var dfitype = typeof(DokanFileInfo);
var dfiMockObject = (DokanFileInfo)FormatterServices.
                    GetUninitializedObject(typeof(DokanFileInfo));
var piofield = dfitype.GetField("_pagingIo",
                    GetField|NonPublic|Instance);
piofield.SetValue(dfiMockObject,true); // Set PagingIo true
var syncIofield = dfitype.GetField("_synchronousIo",
                    GetField|NonPublic|Instance);
syncIoField.SetValue(dfiMockObject,true); // Set SynchronousIo true
var noCacheField = dfitype.GetField("_noCache",
                    GetField|NonPublic|Instance);
noCacheField.SetValue(dfiMockObject,true); // Set NoCache true
var w2eofField = dfitype.GetField("_writeToEndOfFile",
                    GetField|NonPublic|Instance);
[...]

PR #184, opened 2018-05-23
Add public constructor and test support to sealed class DokanFileInfo #183

kyanha:
#183 This is my take on what I'd like to see DokanFileInfo look like, if it's going to remain sealed.
A public constructor that forces it into test mode, test-mode accessors, and prevention of
improper state meddling when it's not constructed for testing.

### DIFF
--- a/DokanNet/DokanFileInfo.cs
+++ b/DokanNet/DokanFileInfo.cs
@@ -46,11 +46,25 @@ namespace DokanNet
         [MarshalAs(UnmanagedType.U1)] private bool _writeToEndOfFile;
 
         /// <summary>
+        /// Determine if this instance was constructed for testing, or by the kernel.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


kyanha:
I'm not entirely certain why the tests are failing (I'm thinking that it's because it's a marshaled block of memory passed from the kernel, and C# likely attempts to use the next bytes of memory for the new bool I put in, which probably segfaults)... but it's really difficult to make non-invasive changes to it.

Please, can DokanFileInfo derive from an interface?  Or, if it can't, can it be unsealed so we can derive test objects from it (preferably with GetRequestor and TryResetTimeout made virtual so they can be overridden)?  Both would alter binary compatibility, I think, but because we're supposed to distribute the DokanNet.dll that we compiled against I don't really think that would be that much of a problem.

PR #188, opened 2018-06-21
Mirror parameters

TrabacchinLuigi:
Added some convenient parameters to run mirror in different folders

### DIFF
--- a/sample/DokanNetMirror/Mirror.cs
+++ b/sample/DokanNetMirror/Mirror.cs
@@ -202,6 +202,12 @@ namespace DokanNetMirror
                 }
                 catch (UnauthorizedAccessException) // don't have access rights
                 {
+                    if (info.Context is FileStream fileStream)
+                    {
+                        // returning AccessDenied cleanup and close won't be called
+                        fileStream.Dispose();
+                        info.Context = null;
+                    }
                     return Trace(nameof(CreateFile), fileName, info, access, share, mode, options, attributes,
                         DokanResult.AccessDenied);
                 }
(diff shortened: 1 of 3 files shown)


Liryna:
Hi @TrabacchinLuigi ,

Thank you for your contribution as always !
This mirror options are well welcomed ! Could you just change the typo imput to input ?

Also it seems like the test CI does not pass 😢 
https://ci.appveyor.com/project/Liryna/dokan-dotnet/build/1.1.0-396/job/wkiewemxkjerrfs4#L293

TrabacchinLuigi:
Hey!
Uhm i didn't even mean to try  to push that commit, wanted to push only the parameters because seemed comfy…
Btw that fixed some files being hanged open in our implementation, maybe setting it to an object instead of null is more the way to go in this implementation, let me see

TrabacchinLuigi:
Ohhh ok i got it... since it was hardcoded that it was mounting C:\ in N:\ the test are not running because it's expecting it to be like that, we need to pass the new parameters for it to work, where can i edit that ?

Liryna:
Humm probably should set the previous value as default if nothing is given in parameters. 
Otherwise `info.Context = null` would be the correct way here. object is only used to flag that it is a folder.

TrabacchinLuigi:
> Humm probably should set the previous value as default if nothing is given in parameters
Gonna lose all the nice interactivity where if no parameters are given user is asked, also someone could want to run tests in different locations, so making the parameters explicit would be easier to understand for them how to change them
> Otherwise info.Context = null would be the correct way here. object is only used to flag that it is a folder.
strange the commit where i give an object is passing AppVeyor again ...

PR #190, opened 2018-06-23
Include the Dokan HResult in the exception msg. This makes diagnosing…

pweerd:
Hi,

Can you please review this change and eventually include it?

Kind regards,
Peter

### DIFF
--- a/DokanNet/DokanException.cs
+++ b/DokanNet/DokanException.cs
@@ -17,7 +17,7 @@ namespace DokanNet
         /// <param name="message">
         /// The error message.
         /// </param>
-        internal DokanException(int status, string message) : base(message)
+        internal DokanException(int status, string message) : base(String.Format ("{0}. (Dokan hr=0X{1:X})", message, status))
         {
             HResult = status;
         }


Liryna:
Hi @pweerd ,

You can directly use `Marshal.GetHRForException` on previous framework when you catch the exception on your side if your goal is to get the hresult.
https://msdn.microsoft.com/fr-fr/library/system.runtime.interopservices.marshal.gethrforexception(v=vs.110).aspx

pweerd:
Hi Liryna,

That is a fast response! Thanks.

Yes, I know, but the downside is that one need to do that while handling the exception.
I wasn't aware, for instance, that the HR was stored in the exception.
I got a generic 'Unknown Error' and needed a few hours to see what the reason was.
Learned a lot about Dokan btw, but having seen the HR would have point me to the correct solution.

So, yes, indeed I modified my application to show the HR, but to me it makes more sense to show the HR in the exception msg by default. (Helping people that are unaware of the presence of the HR in the DokanException)

pweerd:
Hi Liryna ,

No, sorry, it does not help.
My point is that in case of an unknown error, we show "unknown error".
To make this error useful, we should show the NTSTATUS-code in the 
message as well.
Passing it via an implicit or explicit property in the exception makes 
no difference. It forces the caller to catch the exception and raise it 
own exception with the NTSTATUS in the text...

If you wish I can create another PR on top of the PR you talk about.

/P

Liryna:
@pweerd OK. So sorry if I miss your point, you would like to have the error code in the message by default? Because you think people would miss this information if they are not looking at the dokan exception class?

pweerd:
Hi Liryna ,

No problem!
Yes, I would like to see the error code in the message.
The point is that the "unknow error" is basically a catchall. So, if you 
don't know the NTSTATUS, you know nothing.
Looking at the DokanException only tells you that it is a catchall error 
msg.

So, you can guess what was wrong, or, you can recompile your program, 
fetch the NTSTATUS from the exception and show it.
I had to do that because of a DOKAN programming error, which handled 
success as an error, showing me "unknown error" :-)
If I had seen "Unknown error NTStatus=0x00000000", it would have been 
very useful.

/P

lumip:
I might be getting this wrong but I can't really see any benefit.. First of all, the status code passed into the Exception is not a NTStatus value but one of the predefined Dokan error codes. Just exposing this in the exception message does not give you any additional information.. with the message "Unknown Error" it is by definition an error that is not meaningfully represented by any error code. (Actually, currently it should only occur if DokanMain returns 0, i.e. Success status code, which should not raise a DokanException anyways.

>  It forces the caller to catch the exception and raise it own exception with the NTSTATUS in the text...

It would make sense to catch occuring exceptions from libraries in your own code anyways, would it not?

> So, you can guess what was wrong, or, you can recompile your program, fetch the NTSTATUS from the exception and show it. I had to do that because of a DOKAN programming error, which handled success as an error, showing me "unknown error" :-)

Could you maybe go into more detail about how/where this error occurred? It seems like this should be dealt with if it's a programming error in Dokan.

So, to sum up, I can't see what additional information could be obtained from including the Dokan status code in the exception message and would argue in favor of keeping the messages simple.

Liryna:
I agree with @lumip , the error codes are not `NTSTATUS` and all values are known.

pweerd:
I don't know how to explain things better. I'll try for the very last time, because its not worth such a big discussion.

Sorry for mixing up NTSTATUS and DOKAN error codes. That sure was a source of confusion.

I don't agree that all DOKAN errors are known. They are defined in different libraries and so, if 1 component adds a code, it could be unknown to the other component.
Secondly, if the errors are considered all known, why do we have a default case throwing an "unknown error"?
So I really think we should be prepared to handle "unknown" DOKAN errors.

To give you a use case:
I suffered from the problem that was fixed by commit 6707dfd8e59ea57f73a23e9095ffc0399c4c118f
Indeed a dokan programming error, but very difficult to get a clue, because it (well, everything) resulted in the  "unknown error" exception.

"It would make sense to catch occuring exceptions from libraries in your own code anyways, would it not?"
No, not really. I prefer only catching exceptions if I know how to deal with them. So most of the exceptions run through an unhandled exception handler, which simply shows the carried msg.

But anyway, if you prefer keeping messages simple, I have to live with that (and I will) :-)
I will close the pr.

Liryna:
Oh I got your point now🔥 🚀 😎 🚀 🔥 
I am ok to add the error code only for the ErrorUnknown type. This can now easily be done in `GetStatusErrorMessage` 👍 
https://github.com/dokan-dev/dokan-dotnet/blob/cfb279f7a68acec64dcdcebb0004fc2b6edd8178/DokanNet/DokanException.cs

pweerd:
Sure!

PR #193, opened 2018-07-13
Expose strongly typed error status in DokanException

jnagykuhlen:
Currently, the result of DokanMain in case of error is accessible as the HResult of DokanMessage. The meaning of the different possible integer values, however, are not exposed towards the user and only privately listed in the Dokan class as constants.

I would propose to include a public enum DokanStatus in the project that clearly communicates different error states to the user. In addition to the HResult, this strongly typed status value is accessible via the ErrorStatus property in DokanException to allow straightforward error handling in user code.

### DIFF
--- a/DokanNet/Dokan.cs
+++ b/DokanNet/Dokan.cs
@@ -19,62 +19,7 @@ namespace DokanNet
         private const ushort DOKAN_VERSION = 100;
 
         #endregion Dokan Driver Options
-
-        #region Dokan Driver Errors
-        /**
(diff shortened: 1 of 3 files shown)


Liryna:
Hi @jnagykuhlen ,

Thank you for the PR ! very well appreciate and nothing to say much elegant way to handle this 👍

jnagykuhlen:
Hi @Liryna , thanks a lot for the fast response and for incorporating the changes into the project. Keep up the great work with Dokany! :)

PR #197, opened 2018-07-24
Added NotADirectory enum value to DokanResult.

lumip:
NotADirectory was part of NtStatus but not of DokanResult. As it is a valid return value for CreateFile in some cases, it should be included in DokanResult so that consistent uses of enums can be achieved in CreateFile.

I also replaced all usages of NtStatus in the DokanNetMirror implementation with DokanResult for consistency.


Liryna:
Hi @lumip ,

Thank you for your contribution that very estimate ! 🏆

Issue #202, opened 2018-09-11
Marshalling of byte[] arrays is causing excessive GC load (and gen2 collections)


hans-olav:
My app was serving a large file (~2 GB) through Dokan and I was noticing a gen2 garbage collection every second. This is obviously not good for performance - I was seeing 30% of CPU spent in GC. After investigating, it turns out the allocations were coming from marshaling of the byte[] array in ReadFile(). A managed array was being allocated for every ReadFile() call and thrown away after the call completes.

Today's code looks like this:
```
public delegate NtStatus ReadFileDelegate(
    [MarshalAs(UnmanagedType.LPWStr)] string rawFileName,
    [MarshalAs(UnmanagedType.LPArray, SizeParamIndex = 2), Out] byte[] rawBuffer,
    uint rawBufferLength,
...
);
```

By modifying it to not marshal at all and just pass the `IntPtr` I was able to completely eliminate the GC load. The implementor of `IDokanOperations` can then use `Marshal.Copy` or unsafe code to fill the output.

This is obviously a breaking interface change, but maybe it's worth adding a new interface, something like `IDokanOperationsUnsafe`, for those that want the perf?

Liryna:
Hi @hans-olav ,

I see the situation here. Does calling GC Collect to force the previous buffer to be release improve the situation ?

hans-olav:
I haven't tried but I don't think it would help. The file buffers for ReadFile() are usually 1 MB for large files, which means they get allocated directly on the large object heap (LOH) instead of on the gen0 heap - the threshold for that is 85K. Only full GC (gen2 + LOH) collections collect the LOH, and those full GC runs are expensive. In my case, I was serving ~300 MB/s through Dokan, which means marshalling was allocating and throwing away 300 MB of memory every second. This is why GC was kicking in so often and burning 30% CPU.

In other words, the problem isn't that GC isn't happening - it's that full GC is _needed_ that often to keep the high allocation rate from consuming all memory. Calling GC.Collect shouldn't change that - lowering the allocation rate would.

Liryna:
@hans-olav that’s a great answer !
Also you changes would not be an API break if the marshal is done in ReadFileProxy for the user.

Would it be possible for you to make a pull request with the changes ? This would be highly appreciated !

hans-olav:
For the most effective solution, we would need to modify the user interface (`IDokanOperations`) as well, so we should discuss how we would want to do this. There is two levels to this optimization:

[...]

PR #206, opened 2018-09-27
Improving Buffer Handling to Reduce GC Pressure

hans-olav:
**Background**
This is a fix for the issue discussed in #202. The default marshaller for `byte[]` arrays allocates new managed buffers for every native -> managed transition of `Read/WriteFile`. Since these buffers are typically bigger than 85K, they are allocated on the large object heap (LOH) and requires full, expensive Gen2 garbage collection to collect. In some tests, I saw LOH allocation rates of 500+ MB/s and 30+% CPU eaten up by GC - not good.

**The Improvement**
By manually marshalling the unmanaged buffers to managed `byte[]` arrays, we can pool/reuse the managed buffers and get away with almost no LOH allocations. This change implements that by introducing a simple buffer pool that is utilized from `Read/WriteFile`. This is fully backwards compatible with existing applications - the full benefit is realized by using the latest Dokan-DotNet version.

Here is a comparison of copying out a 1 GB file through DokanNetMirror before and after the change.

**Test Setup/Execution**
```
C:\DokanGCTest>fsutil file createnew 1GBfile.dat 1073741824
C:\DokanGCTest>DokanNetMirror.exe
C:\DokanGCTest>copy N:\DokanGCTest\1GBfile.dat .\copytarget.dat /Y
```

**Before**
341 full Gen2 collections to copy the 1 GB file:
image

**After**
0 gen2 collections (even 0 gen0 collections!) to copy the 1 GB file. You may observe that the CPU% isn't all that different between these two samples. This is because there are very few live objects on the managed heap in this test. If Dokan was running in a larger server application with lots of live objects on the heap, the difference would be much larger.
image

**Further Optimization**
This also opens the door to one further optimization opportunity: If the application is reading/writing the file contents from/to an unmanaged buffer (such as a memory mapped file), there is no reason to marshall the data from unmanaged, to managed and back to unmanaged. If Dokan exposed the unmanaged `IntPtr `to the file system implementation, it can use `Buffer.MemoryCopy` to perform the copy directly between two unmanaged buffers.

Since this requires application changes, we expose this new capability as a new sub interface of `IDokanOperations`: `IDokanOperationsUnsafe`. It introduces new overloads to `Read/WriteFile` that takes `(IntPtr, bufferLength)` instead of `(byte[])`. The application can optionally implement this interface to take advantage of the raw buffers.

The PR also includes changes to tests to cover the new functionality:
- Changes to test infrastructure to mount two drives during testing; one for `IDokanOperations `and one for `IDokanOperationsUnsafe`. Tests select during `Setup()` which one they would like to run against.
- A new test class for the new unsafe overloads, that reuses all the tests from `FileInfoTests `by running them in the new mode with unmanaged buffers.
- Upgraded MSTest reference versions and converted `DirectoryInfoTests `to use `DataTestMethod`/`DynamicData `instead of the old `DataSource`/XML file because these tests appeared to be flaky in Appveyor. The new way is neater too.

This contribution from Microsoft is licensed under the MIT license.


js69:
Thanks for this excellent contribution!

Liryna:
Hi @hans-olav ,

Thanks a lot for this beautiful contribution and specially on the explanation and documentation !
Before merging it, could you rebase your commit ?

I know it is probably too much to ask, if you have time, can you add the usage of the new interface optional by a define in the mirror ?
This would be helpful as an example.

Liryna:
Awesome @hans-olav ! Thanks a lot again. This is very appreciated 👍 🏆

PR #220, opened 2019-02-22
Avoid console logger to switch color and never go back

TrabacchinLuigi:
Avoid console logger to switch color and never go back to the original due to multi threading

### DIFF
--- a/DokanNet/Logging/ConsoleLogger.cs
+++ b/DokanNet/Logging/ConsoleLogger.cs
@@ -8,6 +8,7 @@ namespace DokanNet.Logging
     public class ConsoleLogger : ILogger
     {
         private readonly string _loggerName;
+        private readonly object _colorLock = new object();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ConsoleLogger"/> class.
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @TrabacchinLuigi ,

Well found ! Thank you even more for the PR 👍

Liryna:
Humm, just a thought, could be not just change the color at the logger init and change it back during the dispose ? This would speedup and probably give a better result.

TrabacchinLuigi:
the only way to speed this up i've found is to make ProducerConsumer pattern with many threads filling the print queue and a single consumer printing the messages, but it felt overcomplicated for just an example, i can share that implementation if you want, the color have to be saved and restored every time, but like that prucers won't have to wait for the actual print to happen

Liryna:
Just save the color console at logger init and restore during dispose could not just be enough ?

TrabacchinLuigi:
i don't think you createa new instance of the logger every call.. so no, it won't work

Liryna:
What do you mean by creating at every call ?
Logger instance is a Mount param, there is only one during all the process.

TrabacchinLuigi:
Exactly there is only one instance of the logger, so if you don't reset the console color every time, whatever would call `Console.Writeline()` would write in the last used color, and this is not desirable ...

Liryna:
Our color change is already not atomic, the not desirable behavior can already happen between each reset.
We can probably ensure that the console is ours during the process time.

TrabacchinLuigi:
In fact someone could still call `Console.ForegroundColor = something;` making everything vain... so i see what you mean... we could try ... but i don't think this would really speed up much, the big difference is the lock in my opinion

Liryna:
`_colorLock` would not longer exist at this moment.

TrabacchinLuigi:
nope, if you remove `_colorLock` you could end up writing with wrong colors... 
Two threads could enter simultaneously both set the color they want (let's say gray and red) and then both write in red

PR #221, opened 2019-02-22
CreateFile should not check if a path is a file or a directory if the path do not exist

TrabacchinLuigi:
fix: while invoking DIR from command prompt Mirror.CreateFile() was throwing an exception

### DIFF
--- a/sample/DokanNetMirror/Mirror.cs
+++ b/sample/DokanNetMirror/Mirror.cs
@@ -138,7 +138,7 @@ namespace DokanNetMirror
                 try
                 {
                     pathExists = (Directory.Exists(filePath) || File.Exists(filePath));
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @TrabacchinLuigi ,

Thank you for the PR ! Just to be sure, the exception would be a silence now ? did not disturb the mirror behavior ?

TrabacchinLuigi:
Hi, i'm not really sure about what you mean with _"the exception would be a silence now ?"_ 
I can say, i notice that using DIR from command prompt was throwing an exception because it tried to invoke `CreateFile` with weird paths like _N:\whatever\*_ and then to check if it was a directory it was throwing because there was no such path, so it looked safe to assume if the path does not exist, it was not a directory, the exception gone away, and the everything seemed to work the same

Liryna:
Oh Ok I see 👍 All good for me so. Thank you for the details.

PR #222, opened 2019-02-22
Fix: File.SetAttribute could fail, but the file stream will be kept open

TrabacchinLuigi:


Liryna:
Hi @TrabacchinLuigi ,

Thank you for the PR 👍 Well found for the context leak here

TrabacchinLuigi:
What about this not passing some tests ? it looked weird...

PR #223, opened 2019-02-22
Mirror parameters

TrabacchinLuigi:
It just seems i'm not be able to do just one commit

### DIFF
--- a/sample/DokanNetMirror/Program.cs
+++ b/sample/DokanNetMirror/Program.cs
@@ -1,19 +1,35 @@
 using System;
+using System.Linq;
 using DokanNet;
 
 namespace DokanNetMirror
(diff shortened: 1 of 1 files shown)


TrabacchinLuigi:
i'll try to redo this in just one commit, i just don't know GIT so well...

Liryna:
Hi @TrabacchinLuigi ,

Will need to make a rebase your commit see https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git and push force to update github

Otherwise, you seems to have a double mount here https://github.com/dokan-dev/dokan-dotnet/pull/223/files#diff-57540511365d3239e7e88f48c8a938cfR32

PR #224, opened 2019-02-25
Allow mirror to accept parameters to specify what to mount and where

TrabacchinLuigi:


Liryna:
Hi @TrabacchinLuigi ,

Thank you for the contribution 👍 💘  This had to be done much before and waited from others 😄 !

PR #225, opened 2019-02-25
Faster ConsoleLogger

TrabacchinLuigi:
This is a faster version, no threads wait for the others to complete, just one thread write the output


Liryna:
Hi @TrabacchinLuigi ,

This looks to be a good way to handle it 👍 Thanks a lot for your contribution 🏆

Issue #234, opened 2019-08-24
Update wrapper to 1.3.0 Dokan


Liryna:
Dokan 1.3.0 add functions to notify Dokan Kernel that files in user fs has changed DokanNotifyCreate / DokanNotifyDelete / DokanNotifyUpdate / DokanNotifyXAttrUpdate / DokanNotifyRename
Dokan-dotnet should expose this feature
https://github.com/dokan-dev/dokany/releases/tag/v1.3.0.1000

Liryna:
Thanks to @kyanha PR https://github.com/dokan-dev/dokan-dotnet/pull/236

PR #236, opened 2019-10-04
Add methods to support Dokany 1.3.0.1000's new kernel notifications

kyanha:
Added methods to support dokany v1.3.0.1000's new kernel notifications, per #234.

The public dokannet interface for these requires passing the DokanFileInfo that the DokanOperations methods receive, so that the filenames passed can be in the same form that the DokanOperations methods receive (where the root of the filesystem implemented is `@"\"`). This contrasts with the kernel notification functions, which require the full path (including the mountpoint) being part of the filename they're called with.  This means that we need to dig the mountpoint out of the DFI, since implementation classes don't usually track their own mountpoint. We do this with a new internal accessor DokanFileInfo.NativeDokanOptions that exposes the private member DokanFileInfo._dokan_context (which is an IntPtr that points to a DOKAN_OPTIONS structure, and thus not directly useful), so that the MountPoint member can be referenced from it.

### DIFF
--- a/DokanNet/Dokan.cs
+++ b/DokanNet/Dokan.cs
@@ -236,8 +236,11 @@ namespace DokanNet
         {
             StringBuilder FilePathBuilder = new StringBuilder();
             FilePathBuilder.Append(dfi.NativeDokanOptions.MountPoint);
-            // Remove the trailing backslash from the MountPoint, because FilePath includes a leading backslash
-            FilePathBuilder.Remove(FilePathBuilder.Length - 1, 1);
+            if (FilePathBuilder[FilePathBuilder.Length - 1] == '\\')
+            {
+                // Remove the trailing backslash from the MountPoint, because FilePath includes a leading backslash
+                FilePathBuilder.Remove(FilePathBuilder.Length - 1, 1);
+            }
             FilePathBuilder.Append(FilePath);
             return FilePathBuilder.ToString();
         }
--- a/DokanNet/Dokan.cs
+++ b/DokanNet/Dokan.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Text;
 using DokanNet.Logging;
 using DokanNet.Native;
 
@@ -15,7 +16,7 @@ namespace DokanNet
         /// The %Dokan version that DokanNet is compatible with. Currently it is version 1.0.0.
         /// </summary>
         /// <see cref="DOKAN_OPTIONS.Version"/>
-        private const ushort DOKAN_VERSION = 120;
+        private const ushort DOKAN_VERSION = 130;
 
         #endregion Dokan Driver Options
         
@@ -192,7 +193,9 @@ namespace DokanNet
 
             DokanStatus status = (DokanStatus)NativeMethods.DokanMain(ref dokanOptions, ref dokanOperations);
             if (status != DokanStatus.Success)
+            {
                 throw new DokanException(status);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


kyanha:
I should point out that all of the AppVeyor failures are preexisting, and match what I currently get when I try to run tests against the code base.

kyanha:
@Liryna Please note that I'm about to make another PR (to implement IDokanFileInfo and MockDokanFileInfo for test-driven development mocking purposes).  If you accept that one, then this one should be rejected.

If you accept the other, the interface to the notification functions should go into IDokanFileInfo (and be called by the implementation of DokanFileInfo, but mocked by MockDokanFileInfo).  This will make IDokanFileInfo be the only place where kernel callbacks need to be made, to allow for various testing of 'true' and 'false' returns of these functions.  .TryResetTimeout() and .GetRequestor() are already in there, and it makes sense to continue to put file-specific kernel communication in that interface.

Liryna:
Hi @kyanha ,

Thank you very much for adding the 1.3.0.1000 methods ! Very appreciated
Well done with `NativeDokanOptions` 👍 seems to be the best way to do it

Yeah the appveyor test should be fixed. They have been offen fixed 😄 The issue is if appveyor add/update an app that love to access new device in their env, it will make some test fail since we did not expected this operation....
Oh and python seems not to be properly detected now.

I will create issues for this.

Liryna:
`DokanNotify` are not "really" called during operations. Everything happens in the kernel side automatically. `DokanNotify` are just way to force the notification outside all those operations.

Liryna:
@kyanha am sorry, we might need to remove `DokanFileInfo dfi`  as @Lukas0610 said 😢

kyanha:
@

> 
> 
> @kyanha am sorry, we might need to remove `DokanFileInfo dfi` as @Lukas0610 said 😢

Alright, no worries.  I'm sorry for misunderstanding how the functions were supposed to be used in the first place. :)

kyanha:
@Liryna Please see #240 for the removal of dfi from those calls (and NativeDokanOptions from DokanFileInfo).

Lukas0610:
@Liryna Maybe the documentations of the notify-functions should be revised to prevent such misunderstandings from happening again, I doubt this will be a one-time occurrence.

Neither the detailed nor the brief descriptions like "Notify dokan that a file or a directory has been created" on DokanNotifyCreate really explains how it is supposed to be used correctly, adding like "This function has to be called from an outside of the dokan operations, e.g. a file-system change watcher or another event independent of the dokan operations".

Plus updating the mirror-sample with a FileSystemWatcher could also show how it is supposed to be used.

Liryna:
Thank you @kyanha for the quick PR
@Lukas0610 you are right I will look to improve the documentation in dokan.h

Humm mirror sample could have a FileSystemWatcher yes 👍 that's a good idea. I will create a feature request issue for it :D !

Liryna:
@Lukas0610 have you seen https://github.com/dokan-dev/dokany/blob/master/dokan/dokan.h#L893-L901 ?

Lukas0610:
Seems like I overlooked that one, but its still kinda hard to understand, "User FileSystem can notify" sounds like the actual file system implementation is the notifier, just like in this case.

Maybe something like that would make it absolutely clear how the functions should be used: 
_The application implementing the user file system can notify the Dokan kernel driver of external changes.
For example, the mirror application can notify the driver about changes made in the mirrored directory so that those changes will be automatically reflected in the implemented mirror file system_

Liryna:
Good point @Lukas0610 your version is clearer from my point of view too ! Do you think you can do a small PR for this change ? (as it is your idea 😃 )

Lukas0610:
Sure, no problem

Lukas0610:
Just noticed that the `EnableNotificationAPI` flag is missing from `DokanOptions`, like that: https://github.com/Lukas0610/dokan-dotnet/commit/373feeb82d65f052e0d2b7d6b35955db68308ea2#diff-22763f793abf7ce62d289080ca729d65R50

PR #238, opened 2019-10-04
Bringing target framework monikers in line with documentation

kyanha:
Per https://docs.microsoft.com/en-us/nuget/reference/target-frameworks ,
the framework monikers that the project has been using have been
incorrect ('net4.0', 'net4.5', and 'net4.6' are really 'net40', 'net45',
and 'net46' respectively).  This changes them in the .csproj files.
In addition, the [InternalsVisibleTo] attributes on DokanNet (so that
DokanNet.Tests can access them) also needed to be changed so that the
newly generated framework names of DokanNet.Tests were referred to.

Also, I get rid of deprecation warnings for `<ProjectLicenseUrl>` and
`<ProjectIconUrl>` (per https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg)
by changing them to `<ProjectLicense>` and `<ProjectIcon>`.  Note that Visual Studio 2019 doesn't yet support the `type="file"` property in `<ProjectLicense>`; I have an open bug for it at https://developercommunity.visualstudio.com/content/problem/760396/the-and-tags-in-csproj-files-do-not-recognize-or-l.html .


Liryna:
Hi @kyanha 

Thank you for the PR 🎉 
Looks like appveyor is no longer able to build. Can you move the image to vs 2019 here please ?
https://github.com/dokan-dev/dokan-dotnet/blob/master/appveyor.yml#L9

Liryna:
@kyanha your PR has more commit than expected seems like 😢 there is a mix of kernel notification PR

kyanha:
> @kyanha your PR has more commit than expected seems like 😢 there is a mix of kernel notification PR

Ugh, okay.  Apparently I was on the wrong branch when I was doing this?  Or maybe github got confused because the moniker patches were also in the kernel notification branch.  Either way, I'll fix this.

kyanha:
Superseded by #243 .

PR #239, opened 2019-10-04
IDokanFileInfo is now the parameter for IDokanOperations(Unsafe)?.

kyanha:
**THIS IS A BREAKING CHANGE.**

All implementations of `IDokanOperations` will need to change their final
parameters to accept `IDokanFileInfo` instead of bare `DokanFileInfo`.

New class: `MockDokanFileInfo` This is accepted by implementations of
`IDokanOperations` and `IDokanOperationsUnsafe` as their final
parameters.  You can set all of the values of the public
interfaces of `IDokanFileInfo` (which itself was extracted from the
publicly accessible fields of `DokanFileInfo`), and specify the
mock behavior of both `.TryResetTimeout()` and `.GetRequestor()`.

To use this, your unit test mock code needs to behave as though it is
Dokan, calling into your `IDokanOperations` implementation with
various parameters and testing the expected behaviors and return
values.  Integration tests with Dokan still need to use something
like `DokanNet.Tests`.


kyanha:
Associated with #183.  Appveyor actually ran on this one, after commit 9f0933a.

Liryna:
Thank you @kyanha for the PR ! very appreciated and I am sure some people will be glad to finally using unittest 👍

Liryna:
Ah I tought I would be able to squash all the commit but looking at the history you also have the commit from https://github.com/dokan-dev/dokan-dotnet/pull/238 Can you recreate the PR with only the IDokanFileInfo changes ?

kyanha:
I put the target moniker changes in #243, so I'll redo this PR without them.

kyanha:
Superseded by #244

PR #241, opened 2019-10-06
Kernel notification documentation

kyanha:
Adding remarks about what the kernel notification functions in dokany 1.3.0.1000 are about.


kyanha:
...I have no idea why so many commits were pulled into this.  I'm closing this, and will reopen it when I figure it out.

PR #243, opened 2019-10-06
Fixing target framework monikers

kyanha:
Per https://docs.microsoft.com/en-us/nuget/reference/target-frameworks ,
the framework monikers that the project has been using have been
incorrect ('net4.0', 'net4.5', and 'net4.6' are really 'net40', 'net45',
and 'net46' respectively). This changes them in the .csproj files.
In addition, the [InternalsVisibleTo] attributes on DokanNet (so that
DokanNet.Tests can access them) also needed to be changed so that the
newly generated framework names of DokanNet.Tests were referred to.

This also changes appveyor.yml to refer to the correct outputs of the
changed monikers, as well as a small bug which caused multiple
directories to be created in the mirrored drive root.

Also, I get rid of deprecation warnings for <ProjectLicenseUrl> and
<ProjectIconUrl> (per https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg)
by changing them to <ProjectLicense> and <ProjectIcon>. Note that Visual Studio 2019 doesn't yet support the type="file" property in <ProjectLicense>; I have an open bug for it at https://developercommunity.visualstudio.com/content/problem/760396/the-and-tags-in-csproj-files-do-not-recognize-or-l.html .

Supersedes #238.


Liryna:
Thank you for your @kyanha 🏆 very appreciated !

PR #244, opened 2019-10-06
IDokanFileInfo is now the final parameter for IDokanOperations(Unsafe).

kyanha:
**THIS IS A BREAKING CHANGE.**

All implementations of `IDokanOperations` will need to change their final
parameters to accept `IDokanFileInfo` instead of bare `DokanFileInfo`.

New class: `MockDokanFileInfo` This is accepted by implementations of
`IDokanOperations` and `IDokanOperationsUnsafe` as their final
parameters.  You can set all of the values of the public
interfaces of `IDokanFileInfo` (which itself was extracted from the
publicly accessible fields of `DokanFileInfo`), and specify the
mock behavior of both `.TryResetTimeout()` and `.GetRequestor()`.

To use this, your mock code needs to behave as though it is
Dokan, calling into your `IDokanOperations` implementation with
various parameters and testing the expected behaviors and return
values.

### DIFF
--- a/DokanNet.Tests/DokanOperationsFixture.cs
+++ b/DokanNet.Tests/DokanOperationsFixture.cs
@@ -24,6 +24,7 @@ namespace DokanNet.Tests
 
             public bool HasUnmatchedInvocations { get; set; }
 
+            #region Delegates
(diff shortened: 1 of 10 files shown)


kyanha:
Supersedes #239

Liryna:
Thank you very much @kyanha for your PR 👍

PR #248, opened 2019-10-16
Fix package license

zivkan:
As per NuGet's docs on [Packing a license expression or a license file](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-a-license-expression-or-a-license-file), the syntax wasn't quite right.

If you run `dotnet pack` on the project then check `obj/[Debug|Release]/DokanNet.1.3.0.nuspec`, you'll see without these changes there is no `<license` element in the `<metadata>` section, but with this change the license is there.

I opted for using the license expression since it allows customers to see what license it is directly from Visual Studio's Package Manager UI and on nuget.org. When using a file, both just show a link and the customer has to click it to see what the license is. I removed the license file from the package since an expression is used, but if you care about the license file being shipped in the file, feel free to undo that change.

Also FYI, by using the non-standard license file name, GitHub can't detect the license. If you rename the file license.txt, GitHub should be able to start reporting your project as using the MIT license.

### DIFF
--- a/DokanNet/DokanNet.csproj
+++ b/DokanNet/DokanNet.csproj
@@ -15,7 +15,7 @@ This is a .NET wrapper over Dokan, and allows you to create your own file system
     <Company>MaximeC</Company>
     <Authors>MaximeC</Authors>
     <PackageProjectUrl>https://dokan-dev.github.io/</PackageProjectUrl>
-    <PackageLicense>license.mit.txt</PackageLicense>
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @zivkan ,

Thank you very much for looking into our issue on Developer Community 🚀 
and specially providing the guidance and PR for our package.

I was not aware of `PackageLicenseExpression`. This make it easier 👍 
Also have renamed the license file in the project as you said.

PR #253, opened 2020-03-08
Update nuget package references + support for .net standard 2.0

ravenpride:
This PR updates the NuGet package references for .NET Standard 1.3 and adds support for .NET Standard 2.0 to reduce the number of dependencies when targetting .NET Core 2.0 or higher.


Liryna:
Hi @ravenpride ,

Thank you very much for the contribution ! 🏆 
Glad we will have .NET Standard 2.0 support for the next release 😎

PR #271, opened 2021-05-29
Add ability to disable calls to ILogger.Debug

hans-olav:
In a high-performance file system, the calls to `ILogger.Debug` accounts for a significant amount of allocations. Even though the logger may ignore the messages, `object[]` arrays are still allocated for each call that have arguments, and every value type argument is boxed causing further allocations.

For production operation, Debug messages are usually not needed, so to avoid these allocations I am adding a `ILogger.DebugEnabled` property that allows the logger to declare whether it wants debug messages. If not, all the debug logging  (and allocations) is skipped.

It is technically a breaking change to add a method to an existing interface that library users may be implementing. But it only affects anybody that has a custom `ILogger `implementation (the provided ones in DokanNet have been updated). But I deem this worth it over adding a new logger interface because it is so simple to update any custom logger implementations with this property.

### DIFF
--- a/DokanNet/BufferPool.cs
+++ b/DokanNet/BufferPool.cs
@@ -95,7 +95,7 @@ namespace DokanNet
             int poolIndex = GetPoolIndex(bufferSize);
             if (poolIndex == -1 || poolIndex >= _pools.Length)
             {
-                logger.Debug($"Buffer size {bufferSize} not power of 2 or too large, returning unpooled buffer.");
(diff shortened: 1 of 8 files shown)


Liryna:
Hi @hans-olav ,

Thanks you for the contribution! Very appreciated.
I was looking for a reason to make a new dokan-dotnet release and looks like you gave me a good one :)

PR #282, opened 2022-01-04
Unmanaged struct keep-alive logic for async mount

LTRData:
* Implemented manual marshalling of structures and wrapped them and original managed objects in SafeBuffer derived classes to keep managed data alive while the native structure is in use.
* DOKAN_OPTIONS and DOKAN_OPERATIONS changed to classes instead of structs so that they do not need as much copying around and it also makes the "keep-alive" logic simpler.
* IDisposable DokanInstance with a Dispose method that cleans up native allocations.
* DokanHandle class that implements SafeHandle logic for Dokan handles.
* Dokan.CloseFileSystem() removed and replaced with DokanInstance.Dispose().
* Added launchSettings.json to .gitignore because that file contains settings that each developer mostly want to have their own settings. It is only used when debugging in Visual Studio.
* Had to remove netstandard1.3 support to make things simpler. I assume netstandard2.0 is old enough these days anyway.
* Added netcoreapp3.1 target framework for test project and some if's around parts that are .NET Framework specific. This makes it easier to test netstandard2.0 library.

### DIFF
--- a/DokanNet.Tests/DokanNet.Tests.csproj
+++ b/DokanNet.Tests/DokanNet.Tests.csproj
@@ -15,7 +15,7 @@
     net4.6              net4.6
     netstandard1.3      <missing>
     -->
-    <TargetFrameworks>net45;net46;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net46</TargetFrameworks>
     
     <!--Add the Target Framework to the output file names. -->
     <AssemblyName>$(MSBuildProjectName).$(TargetFramework)</AssemblyName>
--- a/DokanNet.Tests/DokanNet.Tests.csproj
+++ b/DokanNet.Tests/DokanNet.Tests.csproj
@@ -15,7 +15,7 @@
     net4.6              net4.6
     netstandard1.3      <missing>
     -->
-    <TargetFrameworks>net45;net46</TargetFrameworks>
+    <TargetFrameworks>net45;net46;netcoreapp3.1</TargetFrameworks>
     
     <!--Add the Target Framework to the output file names. -->
     <AssemblyName>$(MSBuildProjectName).$(TargetFramework)</AssemblyName>
--- a/DokanNet/DokanHandle.cs
+++ b/DokanNet/DokanHandle.cs
@@ -1,18 +1,41 @@
-﻿using DokanNet.Native;
+﻿using System.Runtime.InteropServices;
+using DokanNet.Native;
 using Microsoft.Win32.SafeHandles;
 
 namespace DokanNet
 {
+    /// <summary>
+    /// This class wraps a native DOKAN_HANDLE.
+    /// 
+    /// Since this class derives form SafeHandle, it is automatically marshalled as
+    /// the native handle it represents to and from native code in for example P/Invoke
+    /// calls. It also uses reference counting and guaranteed to stay alive during such calls.
+    /// <see cref="SafeHandle"/>
+    /// </summary>
     internal class DokanHandle : SafeHandleZeroOrMinusOneIsInvalid
     {
+        /// <summary>
+        /// Initializes a new empty instance, specifying whether the handle is to be reliably released.
+        /// Used internally by native marshaller and not intended to be used directly from user code.
+        /// </summary>
+        /// <param name="ownsHandle">true to reliably release the handle during the finalization phase; false to prevent
+        /// reliable release (not recommended).</param>
         public DokanHandle(bool ownsHandle) : base(ownsHandle)
         {
         }
 
+        /// <summary>
+        /// Initializes an empty instance. Used internally by native marshaller and
+        /// not intended to be used directly from user code.
+        /// </summary>
         public DokanHandle() : base(ownsHandle: true)
         {
         }
 
+        /// <summary>
+        /// Releases the native DOKAN_HANDLE wrapped by this instance.
+        /// </summary>
+        /// <returns>Always returns true</returns>
         protected override bool ReleaseHandle()
         {
             NativeMethods.DokanCloseHandle(handle);
--- a/DokanNet/DokanHandle.cs
+++ b/DokanNet/DokanHandle.cs
@@ -1,0 +1,22 @@
+﻿using DokanNet.Native;
+using Microsoft.Win32.SafeHandles;
+
+namespace DokanNet
+{
+    internal class DokanHandle : SafeHandleZeroOrMinusOneIsInvalid
+    {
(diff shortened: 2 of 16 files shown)


sample/DokanNetMirror/DokanNetMirror.csproj
```diff
@@ -14,6 +14,10 @@
     <PackageReference Include="StringInterpolationBridgeStrong" Version="0.9.1" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Properties\" />
+  </ItemGroup>
```
Liryna, line 19: This does not looks to be needed.

DokanNet.Tests/DokanNet.Tests.csproj
```diff
@@ -15,7 +15,7 @@
     net4.6              net4.6
     netstandard1.3      <missing>
     -->
-    <TargetFrameworks>net45;net46</TargetFrameworks>
+    <TargetFrameworks>net45;net46;netcoreapp3.1</TargetFrameworks>
```
Liryna, line 18: It looks like the build does not pass on the CI 😢 
https://ci.appveyor.com/project/Liryna/dokan-dotnet/builds/42074445/job/239r9qpoimordkf9

DokanNet/DokanHandle.cs
```diff
@@ -0,0 +1,22 @@
+﻿using DokanNet.Native;
+using Microsoft.Win32.SafeHandles;
+
+namespace DokanNet
+{
+    internal class DokanHandle : SafeHandleZeroOrMinusOneIsInvalid
```
Liryna, line 6: If you have time, it would be great to add some doc here (even a small one) since all the library is documented

DokanNet/NativeStructWrapper.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace DokanNet
+{
+    internal class NativeStructWrapper<T> : SafeBuffer where T : class
```
Liryna, line 8: If you have time, it would be great to add some doc here (even a small one) since all the library is documented

Liryna:
Thanks @LTRData ! Very appreciated that you looked into this so quickly!!!

sample/DokanNetMirror/DokanNetMirror.csproj
LTRData: Oh, that's right. I did not notice that!

DokanNet/DokanHandle.cs
LTRData: Yes, I'll fix some documentation tags!

DokanNet.Tests/DokanNet.Tests.csproj
LTRData: There is something strange going on with the test project. I can run it with "dotnet test" from command line but from within Visual Studio I always get the error message "Unable to read project information for 'DokanNet.Tests': Sequence contains more than one element" and when trying to rebuild it causes error messages about target frameworks not in project.assets.json, but that is most likely just a cause of the original issue. I would think that there is something wrong somewhere in the project files etc but I cannot figure out exactly what that could be.

DokanNet/NativeStructWrapper.cs
LTRData: Just fixed some documentation comments.

DokanNet.Tests/DokanNet.Tests.csproj
Liryna: I have the same error now even if I remember this was working previously.
@kyanha has even reported the issue to Microsoft here https://developercommunity.visualstudio.com/t/error-occurred-while-restoring-nuget-packages-the-3/761782
I am still using Version 16.11.5, on which version are you? maybe the latest one has a fix ?
Otherwise....I was just locally setting one framework and it worked

LTRData: I am using Visual Studio 2022 17.0.4. I get the same errors when I try in 2019 16.11.8.

For now we could of course set just one framework if that helps. In the future it would be great to test more than one framework though. (I have for example ideas about using IDokanOperationsUnsafe with managed Stream objects using Read and Write overloads for Span<byte> which only works in .NET Core and .NET Standard 2.1 and not in .NET Framework.)

Liryna: Thanks for doing the test. Looks like there is not much we can do besides adding a comment to retry this later so :(

Liryna: Thanks for the change! FYI the release CI now fails due to a memory corruption https://ci.appveyor.com/project/Liryna/dokan-dotnet/builds/42080843/job/lnx5iyhcndi7kdni

LTRData: That is interesting and I saw that now too when I retried "dotnet test" on command line. I'll try to figure out what is happening here.

Liryna:
This PR looks fine. The test issue comes from the overlapped tests which one of them assert due to unexpected data.
`dotnet test --filter OverlappedTests`
Otherwise all other test pass.

Thanks @LTRData for your PR! Merging it and I will try to find out what is happening.

LTRData:
Thanks! There are some obvious errors in the tests. For example a handle could be closed before I/O is complete in some cases. Correcting that helped a bit but not completely. There is another issue somewhere that I cannot find the root cause for right now. The implemention passes other file system test cases I have tried so I am pretty sure there is nothing actually wrong, it is just the test cases in the solution that have some problems.

Liryna:
Well the latest CI run just passed https://ci.appveyor.com/project/Liryna/dokan-dotnet/builds/42134826 ... even if locally I still have failures 😶‍🌫️ 
I have been debugging the kernel to see if all data were completed correctly and it is the case.
So Yeah I am on your side that something is wrong with the test but I am not able to point what.

>Correcting that helped a bit but not completely

Could you share the change ?

Liryna:
Oh just found! https://github.com/dokan-dev/dokan-dotnet/commit/4456212fbd536f658cafa346bf478e4b408a0c97 Have you seen any other issues ?

LTRData:
I have not found anything else that looks like it could have any real impact on anything, but the code is a bit difficult to follow. It could probably be worth it to try to write entirely new test cases for overlapped tests. In earlier versions the threading was different in the driver and so I guess it probably happened to work even though something was probably broken with the test cases already.

PR #285, opened 2022-02-16
Avoid GC of delegates in optimized builds

LTRData:
Referring to: https://github.com/dokan-dev/dokan-dotnet/pull/284

This seem to only happen in optimized builds (Release) and some delegates get collected by GC while the call to `DokanMain` is active. This is solved here by calling `GC.KeepAlive` at the end of `Mount` method so that objects stay alive until that last call.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ TestResults
 *.opendb
 .vs
 *.DotSettings
+**/launchSettings.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,19 +6,24 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-## [2.0.0.0] - 2022-01-01
-- Support dokany 2.0.0.
+## [2.0.1.0] - 2022-01-01
+
+### Added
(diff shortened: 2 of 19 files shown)


uholzer:
Using `GC.KeepAlive` like in your commit [b5f4a05](https://github.com/dokan-dev/dokan-dotnet/pull/285/commits/b5f4a057ae376b18d0b1a5ebc136100a5bf73b4d) fixes my problem and would therefore solve #284.

LTRData:
@Liryna I tried to revert back and create a more "clean" pull request but I am not really sure how to that properly. I cannot really get the history to show what I expect. But in any case, looking at what files would be changed by this pull request there are three changes:
https://github.com/dokan-dev/dokan-dotnet/pull/285/files

So, there is (1) an old attempt to skip overlapped tests when testing for .NET Core, there is (2) this workaround for GC on delegates and there is (3) a small bugfix for RegistryFS sample that fixes a problem where unnamed values caused "file" enumeration to fail in registry keys with such values. Hope this is okay anyway!

Liryna:
Thanks @LTRData for your contribution. I have merged the GC and RegistryFS changes!

PR #288, opened 2022-02-23
Deduplicate code of Mount and CreateFileSystem

uholzer:
As requested in https://github.com/dokan-dev/dokan-dotnet/pull/284/#issuecomment-1043381198 , here the introduction of the ConfigureInstance method, which deduplicates code from Mount and CreateFileSystem.

### DIFF
--- a/DokanNet/Dokan.cs
+++ b/DokanNet/Dokan.cs
@@ -168,58 +168,9 @@ namespace DokanNet
             bool singleThread, int version, TimeSpan timeout, string uncName = null, int allocationUnitSize = 512,
             int sectorSize = 512, ILogger logger = null)
         {
-            if (logger == null)
-            {
-#if TRACE
-                logger = new ConsoleLogger("[DokanNet] ");
-#else
-                logger = new NullLogger();
-#endif
-            }
-
-            var dokanOperationProxy = new DokanOperationProxy(operations, logger);
-
-            var dokanOptions = new DOKAN_OPTIONS
-            {
-                Version = (ushort)version,
-                MountPoint = mountPoint,
-                UNCName = string.IsNullOrEmpty(uncName) ? null : uncName,
-                SingleThread = singleThread,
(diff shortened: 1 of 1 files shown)


Liryna:
Hi @uholzer ,
Thanks for the pull request! Looks like there is a mix of content in the commit. Would you be able to take a look at it ?
Like the `SafeBuffer` for DokanMain and a keepalive removed.

uholzer:
Depends on which stance we want to take:

1. Use DokanInstance and SafeBuffer in both cases, as I have done in this pull request. This means that the class NativeMethods needs to take a SafeBuffer and  that we `GC.KeepAlive` the instance.
2. Not change NativeMethods, i.e. keep passing `DOKAN_OPTIONS` and `DOKAN_OPERATIONS` instead of a SafeBuffer to Mount. Additionally, let ConfigureInstance return a `Tuple<DOKAN_OPTIONS, DOKAN_OPERATIONS>` instead of a DokanInstance. Also, the two GC.KeepAlive we have now, remain as they are.

Given your feedback, I assume you like option 2 better. This makes sense in my opinion as it does not change how we invoke DokanMain. I will update the pull request accordingly.

LTRData:
I agree on # 2 as well. It makes native calls unnecessarily complicated to introduce SafeBuffer etc in calls that actually do not need them. I also assume that the entire GC.KeepAlive thing will be unnecessary at some point in the future and that makes it even more unnecessary to manually marshal objects for this particular native call. GC.KeepAlive has never been needed before and it looks like it is only needed in some very specific scenarios right now most probably due to some framework bug.

uholzer:
Option 2 is ready. I made two functions, one for the options and one for the operations. (I gave up on a single function returning a tuple because destructuring is not possible before .NETFramework 4.7 and doing it manually looks ugly.)

Liryna:
Thank you very much @uholzer for the quick change! 🏆

Issue #280, opened 2022-01-02
Dokan 2.0.0 support


Liryna:
Most dokan [2.0.0](https://github.com/dokan-dev/dokany/releases/tag/v2.0.0.2000) API changes were implemented in dokan-dotnet and the test fixed in master but the new async mount function `CreateFileSystem` is still not working due to the garbage collector releasing the allocated 
`DOKAN_OPTIONS` `DOKAN_OPERATIONS` and maybe even `DokanOperationProxy` shortly after the mount which make the application crash since it is still used by the native code.
https://github.com/dokan-dev/dokan-dotnet/blob/master/DokanNet/Dokan.cs#L382-L429
https://github.com/dokan-dev/dokan-dotnet/blob/master/sample/DokanNetMirror/Program.cs#L37-L40

I have been trying to add a reference by creating a `DokanInstance` class but it did not help unfortunately. If someone has an idea or a fix it would be highly appreciated !
Kind ping to @magol

LTRData:
After a quick look there are a couple of issues with the approach as far as I can see. For instance, DOKAN_OPERATIONS is a managed structure (value type containing object references such as delegates), so when passing such a structure by ref to unmanaged code the runtime marshals the managed structure to a temporary native structure with function pointers instead, just like it would do for a class. This native structure only lives in memory during the native call, when it returns the runtime copies fields back to the managed structure and frees the native structure. The problem is not that the managed structure is freed, it is the temporary copy of it for the function call that is freed.

One way to solve this would be to create a value-only structure that can be directly passed to native code and pin it in memory with GCHandle.Alloc. Such a structure needs to contain only value fields such as IntPtr rather than delegate references and we need to be sure that these delegates that we marshal to unmanaged pointers survive too for at least the needed lifetime. I could take a look at the details here and exactly what needs pinning and manual marshalling and hopefully come up with a suggestion!

Liryna:
Fully agree with what you just said !!! Let me know if I can be any help

LTRData:
Just added a pull request with some changes that should fix this.

kyanha:
[...]

PR #295, opened 2022-03-08
moved .net46 to .net 461 and added .net6.0-window in the TFM

TrabacchinLuigi:
added .net6 compatibility, upped .net4.6 to 4.6.1, and removed the need to sign the test assembly


Liryna:
Thanks for the pull request! Looks like there is an issue with the CI  😢 but it does have 4.6.1 https://www.appveyor.com/docs/windows-images-software/
https://ci.appveyor.com/project/Liryna/dokan-dotnet/builds/42829412/job/rijcjmdmx09eff04

TrabacchinLuigi:
it looked like .net 6 was missing, but, it's really not my area of expertise

Liryna:
True! Well found. @TrabacchinLuigi Can you update the appveyor file at the root in your pull request to change 2019 to 2022?

Liryna:
Garbage collector is behaving differently on release. It has proven to be very helpful to have both run.

Looks like it is building but none are succeeding :(

TrabacchinLuigi:
And how is garbage collection tested ? :raised_eyebrow:
Nevermind i'll revert that part

Liryna:
For example when we moved to V2 we had the GC releasing the native memory only on release https://github.com/dokan-dev/dokan-dotnet/issues/280

PR #296, opened 2022-03-09
Moved .net46 to .net 462 and added .net6.0-window in the TFM

TrabacchinLuigi:
Updated appveyor.yml to use vs2022 image and hardcoded TFM needed to be changed


TrabacchinLuigi:
That test works on my machine

Liryna:
Thanks @TrabacchinLuigi ! Yeah the tests are flaky by their nature being too strict. I might take a look at it later.

Issue #294, opened 2022-03-08
Slightly longer files


TrabacchinLuigi:
As i was investigating a bug in our implementation, I've found out that the problem is reproducible also with the mirror.
Probably it's a dokan-net problem or even a dokany...
The simplest way to reproduce this is to use the test code I've wrote:

```csharp
private static void CreateMemoryMappedFile(String path)
{
    var dataLength = 250;
    var padding = 250;
    var paddedLength = dataLength + padding;
    var fileLength = (Environment.ProcessorCount * paddedLength) - padding;
    if (System.IO.File.Exists(path)) System.IO.File.Delete(path);
    using (var mmf = MemoryMappedFile.CreateFromFile(path, System.IO.FileMode.Create, null, fileLength))
    {
        var tasks = Enumerable.Range(0, Environment.ProcessorCount).Select(n => Task.Run(() =>
        {
            var buffer = new byte[dataLength];
            for (int i = 0; i < dataLength - 1; i++)
            {
                buffer[i] = (Byte)(n + 1);
            }
            var position = paddedLength * n;

            using (var accessor = mmf.CreateViewAccessor(position, dataLength))
            {
                accessor.WriteArray(0, buffer, 0, buffer.Length);
            }
        })).ToArray();
        Task.WaitAll(tasks);
    }
    Debug.Assert(new System.IO.FileInfo(path).Length == fileLength);
}
```

this simply use a memory mapped file to write simultaneously different portions of the file, the file size should be precisely the one expected, but it gets rounded to the next power of 2, which makes me and my colleague think that somewhere we receive a buffer length instead of the actual size of the data to write.

This is not reproducible using something like notepad, but it is consistent with any other drive than dokan

Liryna:
The buffer pool is normally not used if the length is not a power of two.
https://github.com/dokan-dev/dokan-dotnet/blob/b4ee9a72e54b357ed938d54d3eecf41cedd5202a/DokanNet/BufferPool.cs#L94-L100

>Probably it's a dokan-net problem or even a dokany...

That's a good question, can you try on the C mirror/memfs if the issue is present ?

When you enable the logs, what write size is requested ?

TrabacchinLuigi:
can't do it right now, but I'm sure i was writing files with a power of two length, they where 6k and 5750 bytes long.
Small files, and the written file resulted being 8191bytes or 0x1FFF, i think tomorrow i can provide more feedback, is the mirror in c still delivered with the setup ?

[...]

PR #298, opened 2022-03-10
Support pagingIo in mirror.net

TrabacchinLuigi:
Little refactor to remove some warnings and tidy up

### DIFF
--- a/sample/DokanNetMirror/NativeMethods.cs
+++ b/sample/DokanNetMirror/NativeMethods.cs
@@ -1,9 +1,11 @@
-﻿using System.Runtime.InteropServices;
+﻿using System;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
 using Microsoft.Win32.SafeHandles;
 
 namespace DokanNetMirror
 {
-    public static class NativeMethods
+    internal static class NativeMethods
     {
         /// <summary>
         /// Sets the date and time that the specified file or directory was created, last accessed, or last modified.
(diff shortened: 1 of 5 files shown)


TrabacchinLuigi:
Should fix #294

TrabacchinLuigi:
also : is really an uint in unsafe mirror ? can really be that long? i mean, it's still 32bit, and makes more sense, but everywhere else an int32 is used

sample/DokanNetMirror/NativeMethods.cs
```diff
@@ -25,5 +27,14 @@ public static class NativeMethods
         [DllImport("kernel32", SetLastError = true)]
         [return: MarshalAs(UnmanagedType.Bool)]
         public static extern bool SetFileTime(SafeFileHandle hFile, ref long lpCreationTime, ref long lpLastAccessTime, ref long lpLastWriteTime);
+        [DllImport("kernel32.dll", SetLastError = true)]
```
Liryna, line 30: return line missing

Liryna (changes requested):

Liryna:
Thanks @TrabacchinLuigi for fixing the issue! Very appreciated 😎 

Just requested a small change before merge if thats not a problem

TrabacchinLuigi:
i've lot of changes right now, if it's not a problem i'll change that on next pull request

TrabacchinLuigi:
also, i was thinking some atributes are really not needed like marshaling boolean to boolean is the default choiche so no need to explicitly specify that

TrabacchinLuigi:
Dang since i've not made a feture branch i can't do another pull request, git is not my bread and butter, TFS is so much easier

Liryna:
Yeah when you work on the same branch due to your second commit based on another one that is also a PR, everything ends up in the same PR. That's fine if they are clean commit like yours 👍 
Thanks for taking care of the issue your reported!

PR #304, opened 2022-03-22
Readability, IoC, Graceful disposing changes

TrabacchinLuigi:
DokanFileInfo: no need to dispose twice the SafeFileHandle using get's lowered to a try-finally block.
DokanFileInfo: More verbose and explicit way of getting a GCHandle and an IntPtr from it.
DokanInstance: Leverage language expressiveness to make clearer what should not be changed, also expose a property to show if it is disposed.
Mirror Notify class is now disposable and check if DokanInstance has been already disposed before making any call.
Mirror and RegistryFs now handle the CancelKeyPress gracefully.


Liryna:
I will take a look at this tomorrow. It looks 🚀

Liryna:
That's a really good cleanup @TrabacchinLuigi ! Good work and thank a lot again for your contributions!

PR #305, opened 2022-03-31
this is just a test i'll close it soon

TrabacchinLuigi:
this is a proposal about the Solid principles i was talking about in the issues, it's not complete nor polished, but shows where i would suggest going


TrabacchinLuigi:
Damn git i forgot to merge the merge of my modifications before doing this, i just wanted to see if the tests where still working.
Anyway this is a proposal abou the Solid principles i was talking about in the issues, it's not complete nor polished, but shows where i would suggest going

TrabacchinLuigi:
@Liryna before i close this i'd like to know if you like this direction

DokanNet/Dokan.cs
```diff
@@ -533,13 +197,52 @@ public static bool XAttrUpdate(DokanInstance dokanInstance, string filePath)
...
+                {
+                    if (disposing)
+                    {
+                        // dispose managed state (managed objects)
+                        DokanHandle?.Dispose();     // This calls DokanCloseHandle and waits for dismount
+                        DokanOptions?.Dispose();    // After that, it is safe to free unmanaged memory
+                        DokanOperations?.Dispose();
+                    }
+                    //  free unmanaged resources
+                    NativeMethods.DokanShutdown();
```
Liryna, line 223: This object is per mount instance but calling `DokanShutdown` on dispose will make all mount of the process to be shutdown.

Liryna:
Yes, you are in the good way! I have not much to say.
I like the way it feels to be a "newer" C# 👍 

Maybe @LTRData would also want to take a look

DokanNet/Dokan.cs
TrabacchinLuigi: i tought there was a counter in the driver that release everything when it reaches zero, but maybe i'm mistaken

LTRData: No it does not work like that. This shutdown method cannot be called by instance cleanup code in this way. It can only be called once when everything Dokan related should be cleaned up.

TrabacchinLuigi: what would happen if two different processes calls that independetly ? is that safe?

TrabacchinLuigi: ok, i think next week i'll be able to work a little bit on this and propose something!

PR #309, opened 2022-04-23
Use fluent API to configure Dokan and use SOLID principles

TrabacchinLuigi:


DokanNet/Dokan.cs
```diff
@@ -416,7 +33,7 @@ public static uint WaitForFileSystemClosed(this DokanInstance dokanInstance, uin
         /// <param name="driveLetter">Driver letter to unmount.</param>
         /// <returns><c>true</c> if device was unmount 
         /// -or- <c>false</c> in case of failure or device not found.</returns>
-        public static bool Unmount(char driveLetter)
+        public bool Unmount(char driveLetter)
```
Liryna, line 36: This and other functions under that are only calling native methods should be staying static no ? Having to create or get access to the single Dokan instance is not very convenient.

DokanNet/Native/DOKAN_OPTIONS.cs
```diff
@@ -8,7 +8,7 @@ namespace DokanNet.Native
     /// <see cref="NativeMethods.DokanMain"/>
     /// <remarks>This is the same structure as <c>PDOKAN_OPTIONS</c> (dokan.h) in the C version of Dokan.</remarks>
     [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode, Pack = 4)]
-    internal sealed class DOKAN_OPTIONS
+    public sealed class DOKAN_OPTIONS
```
Liryna, line 11: This now public due to OptionsConfigurationDelegate ? I guess that ok. Looking at all the fields and you change it is now all "c#" members 👍

Liryna (commented):
Looks good to me 👍 Great work!

DokanNet/Dokan.cs
TrabacchinLuigi: It was deliberate, to enforce the usage of one Dokan instance that calls initialize and shutdown methods on constructor and dispose

DokanNet/Native/DOKAN_OPTIONS.cs
TrabacchinLuigi: Yep, could hide it with a little more work, but I didn't see any advantage

Liryna:
Thanks again @TrabacchinLuigi for your contribution! Very appreciated 👍

PR #327, opened 2023-04-05
Fixed a crash in `ReadFileProxy` when debug logging is enabled

lostmsu:
The call ends up looking like this:

```csharp
logger.Debug("\tContext\t{FileStream, ...}")
```

And the `Debug` interprets it as a format string, and the part of the string between curly braces gets incorrectly interpreted as a placeholder specification.

IMHO, a mature logging library should be used instead like [Microsoft.Extensions.Logging](https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/)

Or at the very least to prevent an issue like this there should be an overload without `params object[] args` which would have been selected instead, and which would take the string as the final message rather than format string.

### DIFF
--- a/DokanNet/DokanOperationProxy.cs
+++ b/DokanNet/DokanOperationProxy.cs
@@ -398,10 +398,10 @@ namespace DokanNet
             {
                 if (logger.DebugEnabled)
                 {
-                    logger.Debug("ReadFileProxy : " + rawFileName);
-                    logger.Debug("\tBufferLength\t" + rawBufferLength);
-                    logger.Debug("\tOffset\t" + rawOffset);
-                    logger.Debug("\tContext\t" + rawFileInfo);
+                    logger.Debug("ReadFileProxy : {0}", rawFileName);
+                    logger.Debug("\tBufferLength\t{0}", rawBufferLength);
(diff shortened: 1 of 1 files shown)


lostmsu:
Workaround until this is released is to not use any of the standard loggers or `DokanNet.Logging.Logger` with `debug` action set to `null`.

Liryna:
Thanks for the pull request @lostmsu ! That makes me think we should migrate all the logs in this file as you did 😢 since it could also happen for other calls. Do you mind doing the change in this pull request ? That would be very appreciated!

lostmsu:
@Liryna I thought of that, and skimmed the file a bit, and the instances that you see in my PR are the only ones I noticed.

lostmsu:
@Liryna  skimmed again, found and fixed 1 other instance.

Liryna:
Awesome @lostmsu ! Thank you very much for finding and providing the fix 🏆 Do you need me to make a release with the fix ?

lostmsu:
No, I am using the workaround for now.

lostmsu:
Lol I just hit this again in a new project :-D

PR #331, opened 2023-05-12
Fixed Mirror.cs FileStream bug

FrankRodelli:
Fixes a bug where certain pieces of software (Libre office/Free office) try to open a filestream with FileMode.CreateNew and FileShare.Read which throws an IO exception.


Liryna:
Thank you @FrankRodelli for reporting the issue and sharing the fix! Very appreciated

PR #340, opened 2023-09-08
WaitForFileSystemClosedAsync implementation

LTRData:
Requires native DokanRegisterWaitForFileSystemClosed in dokany PR #1171


Liryna:
Hi @LTRData ,

Sorry I really missed the notification for this PR...Thanks for adding the C# piece of your contribution! Is that OK if we wait for the native release to be available before we merge this ?

LTRData:
Thanks! Yes that sounds reasonable!

PR #390, opened 2026-04-27
Fix/invalid name check uses basename

kakapo00:
The previous code checked the full fileName (which includes path separators) against GetInvalidPathChars. Semantically, we should only check the file name portion (baseName) and use the stricter GetInvalidFileNameChars which includes :*?"<>| etc.


LTRData:
I made a combination commit https://github.com/dokan-dev/dokan-dotnet/commit/da29acbb5274785d59aaaf2a8ca0a053beafc47f resolving other issues as well at the same time.

